PR #1, opened 2026-04-19
Add GitHub Actions CI workflow

claymcleod:
Adds `.github/workflows/ci.yml` with three jobs—`test`, `clippy` (warnings-as-errors), and `fmt`—that run on pushes and PRs into `main`, pulling the runner's preinstalled stable Rust (refreshed via `rustup update stable`) and caching the target directory through `Swatinem/rust-cache@v2`.

One thing worth flagging up front: `topcoat-macro` doesn't currently compile (cf. `crates/topcoat-macro/src/lib.rs:50`, where `syn::parse_macro_input!` is being handed a `fn` item instead of a `TokenStream`), so `test` and `clippy` will fail on their first run until that's patched—`fmt` should come back clean.


pikaju (approved):
Thanks again!

PR #21, opened 2026-06-12
Exclude non-mutating file access events from the dev file watcher

akoll:
This event triggered an infinite rebuild loop on linux.

### DIFF
--- a/crates/topcoat-cli/src/dev/mod.rs
+++ b/crates/topcoat-cli/src/dev/mod.rs
@@ -41,11 +41,10 @@ impl DevCommand {
         let (tx, mut rx) = mpsc::unbounded_channel::<notify::Result<notify::Event>>();
         let mut watcher = recommended_watcher(move |event: notify::Result<notify::Event>| {
             if let Ok(ev) = &event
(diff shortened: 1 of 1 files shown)


crates/topcoat-cli/src/dev/mod.rs
```diff
@@ -38,9 +38,14 @@ impl DevCommand {
...
-        let mut watcher = recommended_watcher(move |event| {
-            let _ = tx.blocking_send(event);
+        let (tx, mut rx) = mpsc::unbounded_channel::<notify::Result<notify::Event>>();
+        let mut watcher = recommended_watcher(move |event: notify::Result<notify::Event>| {
+            if let Ok(ev) = &event
+                && matches!(ev.kind, EventKind::Access(_))
+            {
+                return;
+            }
+            let _ = tx.send(event);
```
pikaju:
```suggestion
            if let Ok(ev) = &event
                && !matches!(ev.kind, EventKind::Access(_))
            {
                let _ = tx.send(event);
            }
```

pikaju (changes requested):

PR #81, opened 2026-07-11
Share asset resolver

carllerche:
## Summary

- provide the standard asset route resolver through `AssetRouteResolver::default()`
- reuse that resolver when registering router assets

## Why

Non-request renderers need the same asset URL path logic as the HTTP router. Previously, the standard resolver was constructed only inside `RouterBuilderAssetExt::assets`, so other renderers had to duplicate it.

This gives mailers and other renderers a supported way to install the same resolver in their application context.

This PR is stacked on `codex/decouple-render-context` because that branch introduces the reusable application context consumed by the mailer.

## Checks

- `cargo fmt -p topcoat-asset -- --check`
- `cargo test -p topcoat-asset`
- `cargo clippy -p topcoat-asset --all-targets -- -D warnings`

### DIFF
--- a/crates/topcoat-asset/src/resolver.rs
+++ b/crates/topcoat-asset/src/resolver.rs
@@ -1,5 +1,7 @@
 use crate::BundledAsset;
 
+pub(crate) const ASSET_ROUTE_PREFIX: &str = "/_topcoat/assets";
+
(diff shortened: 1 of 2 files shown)


carllerche:
Doesn't sound like this is the right direction.

PR #126, opened 2026-07-18
fix: count browser string length in UTF-8 bytes

srtdog64:
## Summary

- Make browser `len()` count UTF-8 bytes like Rust; JS previously counted UTF-16 code units.
- Add non-ASCII browser tests and CI coverage.

API follow-up: consider explicit `utf16_len()`, `char_count()`, and version-pinned `grapheme_count()` methods.

## Checks

- `yarn build`
- `yarn test`


pikaju:
Thank you. I should've caught that. In the future we should have rigurous testing to check that the Rust and JavaScript code has identical behavior but I think it is a bit too early for that. I undid your testing harness, I think we want something different, not just testing one individual file like that.

PR #139, opened 2026-07-19
fix: handle procedure wire edge cases

### LINKED ISSUE #137: #[procedure] returning Result<()> makes the awaiting client handler throw and silently stop
A `#[procedure]` declared `-> Result<()>` completes on the server (HTTP 200), but awaiting it in a browser-side handler throws `TypeError: Cannot read properties of null (reading 't')` inside the runtime, and everything after the `.await` is silently skipped. Changing the return type to any non-unit type (and discarding the value at the call site) makes the same handler work.

Reproduced on current `main` (6fc6d1b)

## Reproduction

```toml
[dependencies]
topcoat = "0.2.0"
tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
```

```rust
use topcoat::{
    Result,
    runtime::procedure,
    view::view,
};

#[tokio::main]
async fn main() {
    topcoat::start(Router::builder().discover().build()).await.unwrap();
}

#[procedure]
async fn returns_unit(x: f64) -> Result<()> {
    let _ = x;
    Ok(())
}

#[procedure]
async fn returns_f64(x: f64) -> Result<f64> {
    Ok(x)
}

#[page("/")]
async fn home() -> Result {
    view! {
        <!DOCTYPE html>
        <html>
            <head>
                <title>"unit procedure repro"</title>
                topcoat::dev::script()
            </head>
            <body>
                signal unit_resumed = false;
                signal f64_resumed = false;

                <button id="call-unit" @click=$(async |_e| {
                    returns_unit(1.0).await;
                    unit_resumed.set(true);
                })>"call Result<()> procedure"</button>
                <p id="unit-status" :hidden=$(!unit_resumed.get())>
                    "handler resumed after Result<()> procedure"
                </p>

                <button id="call-f64" @click=$(async |_e| {
                    returns_f64(1.0).await;
                    f64_resumed.set(true);
                })>"call Result<f64> procedure"</button>
                <p id="f64-status" :hidden=$(!f64_resumed.get())>
                    "handler resumed after Result<f64> procedure"
                </p>
            </body>
        </html>
    }
}
```

Run with `topcoat dev`, open the page, click both buttons.

**Expected:** both status lines appear.

**Actual:** only the `Result<f64>` line appears. The `Result<()>` click logs

```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 't')
```

and `unit_resumed.set(true)` never runs. The POST itself succeeds, but the page just looks like it ignored the rest of the handler.

### LINKED ISSUE #138: Zero-argument procedures can't be called
A `#[procedure]` with no arguments can never be successfully called from the browser: the client runtime POSTs `[]`, but the generated server handler deserializes the argument tuple as `()`, which serde_json only accepts as `null`. Every call fails with 400, the `.await` in the calling handler throws `Procedure call failed: 400 Bad Request`, and the rest of the handler is silently skipped.

Reproduced on current `main` (6fc6d1b)

## Reproduction

```toml
[dependencies]
topcoat = "0.2.0"
tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
```

```rust
use topcoat::{
    Result,
    runtime::procedure,
    view::view,
};

#[tokio::main]
async fn main() {
    topcoat::start(Router::builder().discover().build()).await.unwrap();
}

#[procedure]
async fn ping() -> Result<f64> {
    Ok(1.0)
}

#[page("/")]
async fn home() -> Result {
    view! {
        <!DOCTYPE html>
        <html>
            <head>
                <title>"zero-arg procedure repro"</title>
                topcoat::dev::script()
            </head>
            <body>
                signal resumed = false;

                <button @click=$(async |_e| {
                    ping().await;
                    resumed.set(true);
                })>"call zero-arg procedure"</button>
                <p :hidden=$(!resumed.get())>"handler resumed"</p>
            </body>
        </html>
    }
}
```

Run with `topcoat dev`, open the page, click the button.

**Expected:** "handler resumed" appears.

**Actual:** it never appears. The Network panel shows the client's request and the server's answer disagreeing about the wire format:

```
POST /_topcoat/procedures/<id>
request body:  []
response:      400  bad request: invalid JSON value: invalid type: sequence, expected unit at line 1 column 0 (at `.`)
```

and the console logs `Uncaught (in promise) Error: Procedure call failed: 400 Bad Request`.

The same mismatch is reproducible with curl: the endpoint accepts `null` (HTTP 200) and rejects the `[]` the client actually sends.

srtdog64:
Fixes #137. Fixes #138.

- Send `null` for zero-argument procedure tuples.
- Hydrate unit responses and Rust's flat `Result` wire format.

Checks: browser build, TypeScript check, focused Biome lint, and temporary bundled assertions.

After testing this end to end, I think the recurring problem is at the Rust-to-JavaScript transpiler I/O boundary. The contract between Rust surrogates, generated JavaScript, serde wire formats, and hydration/dehydration is still implicit, so boundary bugs keep recurring.

tested result:
`zero-argument procedure: 0.0 -> 7
Result<()> continuation: false -> true
nested Result hydration: 0.0 -> 9
browser warnings/errors: []`

This is only my personal view, but while these are individually small bugs, they seem to come from different representations being assumed on each side. I think a canonical wire schema and bidirectional conformance tests could make this boundary safer.


pikaju:
Thank you!

pikaju (approved):

srtdog64:
You understand the project’s design goals better than I do, so please treat this only as feedback from trying it a few times. My concern is that similar bugs may continue to appear while the contract between Rust and JavaScript remains implicit. Personally, I think it may be worth designing that foundation now.
I currently lean toward a JavaScript-oriented wire schema, especially because I found some behavior difficult to use confidently in applications handling non-ASCII and UTF-8 text. However, I understand that your design priorities may be different.
One possible incremental approach could be:
1.Define shared JSON fixtures for unit, tuples, Option, Result, and nested combinations.
2.test Rust serialization and deserialization against those fixtures.
3.Test Javascript hydraiton and dehydration against the same fixtures with Vitest.
4.Once the format is stable, generate TypeScript types from JSON Schema or Rust definitions.
If these four pieces shared one contract, I think the framework would become much easier to test and use reliably. At the moment, it is difficult for me to know whether the same expression will behave identically on the server and in the browser.

p.s Or a neutral protocol design might work too. It doesn't necessarily have to be JS (that's just what I'm more familiar with since I use it more often... it might just be that).

It feels like a single source of truth is needed. Currently, since there's no protocol specification, it's hard to know which one is the 'truth.'

Please understand that I respect you, and I hope you'll bear with me since my English is a bit lacking

PR #147, opened 2026-07-20
feat: concurrent and observable remote asset fetching

rivertam:
A production Docker build spent ~160s in `topcoat asset bundle` printing nothing. The issue was the host failing to fetch the DNS and that taking 5s synchronously (in serial) to fall back to the correct DNS. That's a host issue, but the lack of parallelization in the asset bundler here made a 5s problem into a 160s problem.

I did use Claude to write this, but I'm a real person/programmer I promise. :)

I'm not sure if the callback (`on_fetch`) is a nice API. I would probably favor a stream based API, whether tokio streams or iterators for the sync, but let me know what you think.

### DIFF
--- a/crates/topcoat-cli/src/dev/build.rs
+++ b/crates/topcoat-cli/src/dev/build.rs
@@ -93,7 +93,13 @@ async fn build(kind: BuildKind, opts: BuildOpts) -> Option<PathBuf> {
     };
 
     let spinner = Spinner::new("bundling assets");
-    let bundled = crate::asset::run_bundle(&bytes, None).await;
+    let progress = spinner.bar();
+    let bundled = crate::asset::run_bundle(&bytes, None, move |event| {
+        if let topcoat_asset::FetchEvent::Downloaded { uri, .. } = event {
+            progress.set_message(format!("bundling assets ({uri})"));
+        }
+    })
+    .await;
     drop(spinner);
 
     if let Err(error) = bundled {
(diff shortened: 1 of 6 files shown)


crates/topcoat-cli/src/dev/build.rs
```diff
@@ -93,7 +93,13 @@ async fn build(kind: BuildKind, opts: BuildOpts) -> Option<PathBuf> {
     };
 
     let spinner = Spinner::new("bundling assets");
-    let bundled = crate::asset::run_bundle(&bytes, None).await;
+    let progress = spinner.bar();
+    let bundled = crate::asset::run_bundle(&bytes, None, move |event| {
```
rivertam, line 97: both `run_bundle` and `Bundler::bundle` effectively just take a callback rather than returning an iterator that can be `match`ed upon... I would probably favor the iterator. In the case of `run_bundle` (which has `tokio`), it can return a tokio stream while the sync api would expose just a sync iterator

pikaju:
Hey, thank you. I took inspiration from your PR and created a new one here: #157 

I added a `BundlerConfig` and made it so the whole bundling pipeline is parallel, not just the downloads.

As for the DNS caching, are you sure this is needed? I found this issue on `ureq` where it sounds like the operating system should really do the DNS caching: https://github.com/algesten/ureq/issues/401 

Please let me know if my solution doesn't solve your problem, otherwise let's go with that one for now.

rivertam:
@pikaju thank you! Nope, seems unnecessary. :)

PR #150, opened 2026-07-20
fix: decode padded base64 session tokens correctly

### LINKED ISSUE #149: Token::decode rejects every valid session token
### Summary

`Token::decode` rejects every token produced by `Token::encode`, so no session is ever recognized. `CookieTokenStore::read` does `Token::decode(...).ok()`, which silently becomes `None`, and `refresh`/`rotate`/`stop` then all resolve to `None` for a client presenting the exact token the server issued.

### Location

`crates/topcoat-session/src/token.rs` (`Token::decode`)

### Root cause

`Token::encode` uses the **padded** `URL_SAFE` engine, so a 32-byte token encodes to 44 characters. `decode` uses `decode_slice` into a fixed `[0u8; 32]` buffer:

```rust
let mut bytes = [0u8; 32];
let num_bytes = URL_SAFE
    .decode_slice(s, &mut bytes)
    .map_err(|error| match error {
        DecodeSliceError::OutputSliceTooSmall => DecodeError::Length,
        DecodeSliceError::DecodeError(error) => error.into(),
    })?;
```

`decode_slice` first runs a conservative size pre-flight, `decoded_len_estimate(len) = ceil(len / 4) * 3`, which **ignores padding**. For a 44-char input that is `11 * 3 = 33`. The output buffer is only 32 bytes, so `32 < 33` returns `DecodeSliceError::OutputSliceTooSmall` **before any decoding happens**, even though the true decoded length is exactly 32. That is mapped straight to `DecodeError::Length`. The later `num_bytes != 32` guard never runs on the happy path.

### Reproduction

```rust
let token = Token::random();
let encoded = token.encode();          // 44 chars, padded URL-safe base64
assert!(Token::decode(&encoded).is_ok()); // fails today: Err(DecodeError::Length)
```

There was no encode -> decode round-trip test, so CI stayed green.

### Impact

The session system is non-functional end to end: a freshly issued token never resolves to a session. It fails closed (no false sessions), but no session is ever readable.

### Suggested fix

Decode into a `Vec` and convert, so the padding-aware true length is used:

```rust
use base64::{Engine as _, engine::general_purpose::URL_SAFE};
let bytes = URL_SAFE.decode(s)?;
let bytes: [u8; 32] = bytes.try_into().map_err(|_| DecodeError::Length)?;
Ok(Self::new(bytes))
```

(Alternatively keep `decode_slice` but size the buffer `[0u8; 33]`.)

I have a fix with a round-trip regression test ready to submit.

thejesh23:
Fixes #149.

### Problem

`Token::encode` uses the **padded** `URL_SAFE` engine, so a 32-byte token encodes to 44 characters. `Token::decode` decoded into a fixed `[0u8; 32]` buffer via `decode_slice`, whose size pre-flight `decoded_len_estimate(44) = ceil(44 / 4) * 3 = 33` ignores padding and exceeds the 32-byte buffer. Every valid token was therefore rejected with `DecodeError::Length` **before any decoding ran**, and `CookieTokenStore::read` (`Token::decode(...).ok()`) turned that into `None` — so a client presenting the exact token the server issued never resolved to a session.

### Fix

Decode into a `Vec` and convert to `[u8; 32]`, so the padding-aware true length is used:

```rust
let bytes = URL_SAFE.decode(s)?;
let bytes: [u8; 32] = bytes.try_into().map_err(|_| DecodeError::Length)?;
```

Behavior is unchanged for genuinely invalid input: non-base64 -> `DecodeError::Base64`, wrong length -> `DecodeError::Length`.

### Tests

Added a `#[cfg(test)] mod tests` covering the previously-missing round trip:

- `encode_then_decode_round_trips` — fails on `main`, passes here.
- `decode_rejects_wrong_length`
- `decode_rejects_invalid_base64`

`cargo test -p topcoat-session --all-features`, `cargo clippy -p topcoat-session --all-targets --all-features -- -D warnings`, and `cargo fmt -- --check` all pass locally.

### DIFF
--- a/crates/topcoat-session/src/token.rs
+++ b/crates/topcoat-session/src/token.rs
@@ -36,17 +36,9 @@ impl Token {
     /// Returns a [`DecodeError`] when `s` is not valid base64 or does not
     /// decode to exactly 32 bytes.
     pub fn decode(s: &str) -> Result<Self, DecodeError> {
-        use base64::{DecodeSliceError, Engine as _, engine::general_purpose::URL_SAFE};
-        let mut bytes = [0u8; 32];
-        let num_bytes = URL_SAFE
-            .decode_slice(s, &mut bytes)
-            .map_err(|error| match error {
-                DecodeSliceError::OutputSliceTooSmall => DecodeError::Length,
-                DecodeSliceError::DecodeError(error) => error.into(),
-            })?;
-        if num_bytes != bytes.len() {
-            return Err(DecodeError::Length);
-        }
+        use base64::{Engine as _, engine::general_purpose::URL_SAFE};
+        let bytes = URL_SAFE.decode(s)?;
+        let bytes: [u8; 32] = bytes.try_into().map_err(|_| DecodeError::Length)?;
         Ok(Self::new(bytes))
     }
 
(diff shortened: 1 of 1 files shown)


pikaju:
See #149

PR #152, opened 2026-07-20
fix: escape signal comment payload to prevent XSS

### LINKED ISSUE #151: Signal value written into HTML comment unescaped -> XSS via -->
### Summary

A `Signal`'s initial value is emitted into an HTML comment marker without any escaping, so a value containing `-->` terminates the comment early and drops the browser back into live HTML parsing. An attacker-controlled (or otherwise untrusted) signal value can inject live HTML/script -- a stored/reflected XSS depending on where the value comes from.

### Location

`crates/topcoat-runtime/src/signal.rs`, `SignalDeclaration::into_view_parts`:

```rust
parts.push_str_unescaped("<!-- ::topcoat::signal(");
parts.push_str_unescaped(json);   // <- untrusted value, written raw
parts.push_str_unescaped(") -->");
```

### Root cause

`json` embeds the signal's runtime value and is written with `push_str_unescaped`, whose own docs say "Use this only for trusted markup. Passing untrusted input defeats the runtime's escaping and can lead to XSS vulnerabilities." `serde_json` escapes only `"`, `\`, and control characters -- not `>` or `-`. So a value containing `-->` closes the comment.

The client reads this marker back with `JSON.parse` verbatim (`crates/topcoat-runtime/browser/src/comment.ts`, the `signal` branch), so whatever escaping is used on the server must remain valid JSON that `JSON.parse` decodes to the same value. (Note: unlike the `expr` and `scope` markers, the `signal` branch does not run `decodeHtml` on its payload.)

### Reproduction

```rust
Signal::new("--><img src=x onerror=alert(1)>".to_string())
```

renders

```html
<!-- ::topcoat::signal({"t":"signal","id":"...","v":"--><img src=x onerror=alert(1)>"}) -->
```

The tokenizer ends the comment at the first `-->`, making `` a live element.

### Impact

XSS wherever a signal's value can contain attacker-influenced data.

### Suggested fix

Escape the HTML-significant characters (`<`, `>`, `&`) in the serialized payload as JSON `\uXXXX` escapes before writing it. This keeps the payload byte-identical after `JSON.parse` on the client (so no client change is needed) while making `-->` unformable:

```rust
fn escape_for_html_comment(json: &str) -> String {
    json.replace('<', "\\u003c")
        .replace('>', "\\u003e")
        .replace('&', "\\u0026")
}
```

(An alternative that mirrors the `expr`/`scope` markers would be to route the payload through `HtmlContext::Comment` on the server and add a `decodeHtml` call to the `signal` branch of `comment.ts`; I went with the server-only `\uXXXX` approach to avoid a coordinated client change.)

I have a fix with regression tests (comment-cannot-terminate + JSON round-trip) ready to submit.

thejesh23:
Fixes #151.

### Problem

`SignalDeclaration::into_view_parts` wrote the signal's serialized value into its HTML comment marker with `push_str_unescaped`. `serde_json` does not escape `>` or `-`, so a signal value containing `-->` closed the comment early and dropped the browser back into live HTML parsing:

```rust
Signal::new("--><img src=x onerror=alert(1)>".to_string())
// -> <!-- ::topcoat::signal({... "v":"--><img src=x onerror=alert(1)>"}) -->
//                                    ^^^ comment ends here; <img ...> is live -> XSS
```

### Fix

Escape the HTML-significant characters (`<`, `>`, `&`) in the serialized payload as JSON `\uXXXX` escapes before writing it:

```rust
fn escape_for_html_comment(json: &str) -> String {
    json.replace('<', "\\u003c")
        .replace('>', "\\u003e")
        .replace('&', "\\u0026")
}
```

Why this is safe and needs no client change: the browser runtime reads this marker back with `JSON.parse` verbatim (`comment.ts`, the `signal` branch does not `decodeHtml`). `\uXXXX` is standard JSON, so `JSON.parse` decodes it back to the exact original character -- the value is byte-identical after the round trip -- while `-->` can no longer appear in the payload. `serde_json` already emits `"` as `\"`, so structural quotes are untouched.

I verified the JS side round-trips the escaped payload:

```
server payload: {"t":"signal","id":"x","v":"--></script> a < b && \"quoted\""}
contains literal "-->": false
JSON.parse(...).v === original value: true
```

### Scope

This targets the `signal` marker, which carries arbitrary runtime values. The `expr`/`scope` markers carry compile-time-generated JS / UUIDs and are `decodeHtml`d on the client, so they are out of scope here.

An alternative, more symmetric with `expr`/`scope`, would be to route the payload through `HtmlContext::Comment` on the server and add `decodeHtml` to the `signal` branch of `comment.ts`. I chose the server-only `\uXXXX` escape to keep the change minimal and avoid a coordinated Rust + TypeScript change; happy to switch if you prefer the symmetric approach.

### Tests

Added a `#[cfg(test)] mod tests` in `signal.rs`:

- `signal_payload_cannot_terminate_the_comment` -- a malicious `-->...` value produces exactly one `-->` (the marker's own close) and no live `<img`. Fails on `main`, passes here.
- `signal_payload_round_trips_through_json` -- the emitted payload is valid JSON that decodes back to the original value unchanged.

`cargo test -p topcoat-runtime --all-features`, `cargo clippy -p topcoat-runtime --all-targets --all-features -- -D warnings`, and `cargo fmt -- --check` all pass locally.

### DIFF
--- a/crates/topcoat-runtime/src/signal.rs
+++ b/crates/topcoat-runtime/src/signal.rs
@@ -100,9 +100,26 @@ where
             .expect("failed to serialize signal declaration payload");
 
         parts.push_str_unescaped("<!-- ::topcoat::signal(");
-        parts.push_str_unescaped(json);
+        parts.push_str_unescaped(escape_for_html_comment(&json));
         parts.push_str_unescaped(") -->");
     }
+}
+
+/// Rewrites the HTML-significant characters in a serialized JSON payload as
+/// JSON `\uXXXX` escapes so the payload cannot terminate the HTML comment it is
+/// embedded in.
+///
+/// The client reads this marker back with `JSON.parse` (see the browser
+/// runtime's `comment.ts`), which decodes `\uXXXX` escapes, so the value is
+/// byte-identical after the round trip. `serde_json` does not escape `<`, `>`,
+/// or `&`, so without this a signal value containing `-->` would close the
+/// comment early and drop the browser back into live HTML parsing -- an XSS
+/// vector. `"` is already emitted as `\"` by `serde_json`, so structural quotes
+/// are left untouched.
+fn escape_for_html_comment(json: &str) -> String {
+    json.replace('<', "\\u003c")
+        .replace('>', "\\u003e")
+        .replace('&', "\\u0026")
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


pikaju:
Thank you for the effort! But I wanted to solve the issue differently, see #154

PR #168, opened 2026-07-23
fix: return Bool surrogates from boolean event fields

AmeinEskinder:
## Problem

The Rust-side `Event` / `EventTarget` surrogates declare every boolean field as `BoolSurrogate`, but the browser runtime returns raw JS booleans for all ten of them (`alt_key`, `bubbles`, `cancelable`, `ctrl_key`, `default_prevented`, `is_composing`, `meta_key`, `repeat`, `shift_key`, and `target.checked`) — while string and number fields are already wrapped in their `String`/`F64` counterparts.

Writing one of these raw booleans into a signal poisons it: the next `.get()` calls `.clone()` on a primitive and throws

```
Uncaught TypeError: this.read(...).deref(...).clone is not a function
```

which kills the page's reactivity.

## Repro

The checkbox pattern from the runtime guide:

```rust
view! {
    signal open = false;
    <input type="checkbox" @change=$(|e: Event| open.set(e.target.checked))>
    <p :hidden=$(!open.get())>"..."</p>
}
```

First click throws the TypeError above; the `:hidden` bind never updates again.

## Fix

- Wrap all ten boolean getters in the `Bool` surrogate (making `boolProp` return `Bool`, matching what `stringProp`/`f64Prop` already do).
- Make `WriteSignal.get()` tolerate non-clonable values the same way `Option`/`Result` already do — defense in depth against any future primitive leak.
- `dist/index.js` rebuilt with `tsup && tsc`; `vitest run` passes.

## Verified

On Windows/Chrome against a test app: the repro above now toggles visibility both directions with no console errors, and the signal survives repeated `e.target.checked` writes.


pikaju (approved):

pikaju:
Thank you!

PR #169, opened 2026-07-23
fix: topcoat dev hot reload never succeeds on Windows (exe file lock)

AmeinEskinder:
## Problem

`topcoat dev` keeps the previous application process serving while a rebuild is in flight — but on Windows a running process locks its image file, so cargo can never replace the executable at the link step. Every rebuild after the first fails with:

```
error: failed to remove file 'target\debug\app.exe'
Caused by: Access is denied. (os error 5)
```

and the watcher settles into `previous build still running; waiting for changes...` permanently. Hot reload never succeeds on Windows; the only way out is killing the dev server and restarting it for every change.

## Fix

`AppServer::spawn` copies the built executable to a shadow path (`app.topcoat-dev.exe`) and runs the copy, leaving the original free for the next link. The copy is retried briefly because Windows can hold the previous image for a moment after the old process is reaped. Non-Windows platforms run the original path unchanged.

The `BuildStamp` up-to-date check is unaffected — it stamps the original path, which cargo continues to own.

## Verified

On Windows 11: `topcoat dev` on an example app, edit a source file, rebuild completes, server restarts, page serves the new markup (`ready on ...` twice in the log where previously the second cycle always failed with os error 5).

### DIFF
--- a/crates/topcoat-cli/src/dev/app_server.rs
+++ b/crates/topcoat-cli/src/dev/app_server.rs
@@ -1,5 +1,5 @@
 use std::io;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::{ExitStatus, Stdio};
 
 use tokio::process::{Child, Command};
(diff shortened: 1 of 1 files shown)


pikaju (approved):

pikaju:
Thank you!

PR #171, opened 2026-07-23
fix: remove double type assertion in DOM binding

Sethispr:
replaced the `(el as unknown as Record<string, unknown>)` double cast with a single intersection cast `(el as Element & Record<string, unknown>)`

this way it preserves the Element type identity instead of blowing it away with unknown


pikaju:
Seems like an unnecessary change but sure

pikaju (approved):

PR #193, opened 2026-07-24
feat(cli): topcoat new CLI command to bootstrap pre-configured projects

JijoBose:
## Summary

Implements the roadmap item "`topcoat new` CLI command to bootstrap
pre-configured projects." Adds a top-level `topcoat new` command that scaffolds
a complete, ready-to-run Topcoat project from a chosen template, so getting
started no longer means hand-assembling `Cargo.toml`, `main.rs`, and a build
script from the docs.

## What's included

- **`topcoat new <path>`** — creates the project directory, derives the package
  name from the path (or `--name`), initializes a repository, and writes the
  template files. Refuses an existing destination and validates the package name
  up front.
- **Three templates**, each grounded in an existing compiled example so they
  stay in sync with the framework:

  | Template  | Scaffolds |
  |-----------|-----------|
  | `minimal` (default) | A single page rendering a component (the getting-started app) |
  | `tailwind` | A Tailwind-styled page with `build.rs` and the `tailwind` feature wired up |
  | `runtime`  | An interactive counter using client-side signals + the runtime asset bundle |

  With no `--template`, an interactive picker prompts for one (and errors with a
  clear message in non-interactive contexts).
- **`--vcs <VCS>`** — the version control system to initialize a repository for,
  taking cargo's own values: `git`, `hg`, `pijul`, `fossil`, or `none`. Defaults
  to `git`, or to no repository when the project lands inside an existing git or
  Mercurial checkout; passing `--vcs` explicitly overrides that, as in cargo.
  Each system gets the ignore file it actually reads (`.gitignore`, `.hgignore`,
  `.ignore`, `.fossil-settings/{ignore,clean}-glob`) with cargo's exact `target`
  patterns, and `none` writes none, since nothing would read it.

  A system named with `--vcs` must work or the command fails with the reason,
  reclaiming the directory it created so a retry isn't blocked. The defaulted
  `git` stays best effort in comparison: on a machine without git the project is
  still scaffolded, just without a repository. The repository is created before
  the files are written, since `fossil open` refuses a non-empty directory.
- **`--path <DIR>`** — depend on a local `topcoat` checkout by path instead of
  crates.io, for testing an unreleased `topcoat`. Accepts the crate directory or
  a workspace root containing `crates/topcoat`; records an absolute path and
  handles the `tailwind` build-dependency's `default-features = false`.
- **Docs** — new `crates/topcoat-cli/docs/new.md` guide, linked from the README
  "Start here" list; roadmap item removed.
- Adds `toml` as a dev-dependency (used to validate generated manifests in tests).

## Testing

- 17 unit tests: package-name validation, path/name inference, symlink-aware
  destination guard, TOML validity of every generated manifest, `syn`-parse of
  every generated `main.rs`/`build.rs`, feature/build-script correctness,
  `--path` crate resolution, the `--vcs` value vocabulary and its ignore files,
  and enclosing-repository detection.
- Manually generated all three templates and confirmed each **compiles and
  serves**: `minimal` -> "Hello, World!"; `runtime` -> working counter with its
  runtime asset resolving; `tailwind` -> styled page with the stylesheet asset
  returning 200.
- Manually exercised the `--vcs` matrix: the default writes `.gitignore` and
  `.git`; `--vcs none` writes neither; inside an existing repository the default
  creates nothing while `--vcs git` creates one anyway; `--vcs bogus` is rejected
  listing the five values; and an explicit `--vcs` naming an uninstalled system
  fails with the reason and leaves nothing behind. Confirmed against real
  `cargo new` that it likewise omits `.gitignore` inside an existing repository,
  so this matches cargo rather than deviating.
- `cargo fmt` and `cargo clippy` clean.

## Note for reviewers

The templates target the framework's current API (e.g. the post-#166 layout
signature `slot: Result`). The crates.io `topcoat` a generated project pins is
whatever version matches the CLI, so once this ships alongside a `topcoat`
release the templates build against crates.io unchanged. To try the
`tailwind`/`runtime` templates from *this branch* before such a release, use
`--path` to point at your local checkout, and create the project outside the
`topcoat` workspace so cargo doesn't absorb it:

```sh
topcoat new my-app -t tailwind --path /path/to/topcoat
cd my-app && topcoat dev
```

Two things worth flagging on the `--vcs` work:

- The `pijul` and `fossil` init paths are untested — neither binary is present in
  CI — so those arms are code review only.
- The `tailwind` template's `build.rs` now scopes the class scan with
  `.cwd("src")`. Scanning the package root relies on `.gitignore` to stay out of
  `target/`, which is not there under `--vcs none` or a non-git system; scoping
  the scan removes the dependency and speeds it up for git users too. Verified
  the generated stylesheet still contains every class from `src/main.rs`.
- There is no config-file default, so cargo's `cargo-new.vcs` has no equivalent
  here.


crates/topcoat-cli/src/new.rs
```diff
@@ -0,0 +1,346 @@
...
+    match selection {
+        Some(index) => Ok(Template::ALL[index]),
+        None => Err("no template selected".to_string()),
+    }
+}
+
+/// Initialize a git repository in `root`, unless git is unavailable or `root`
+/// already sits inside an existing repository (a nested repo would surprise).
+/// Returns whether a repository was created.
+fn init_git(root: &Path) -> bool {
```
atamakahere-git: A suggestion, Supporting more than just `git` would be a good idea
Ref: https://github.com/rust-lang/cargo/blob/0158e40d8638a7de292b7242b1533caaf48cbe5f/src/util/command_prelude.rs#L895

atamakahere-git: Also: https://github.com/rust-lang/cargo/blob/0158e40d8638a7de292b7242b1533caaf48cbe5f/src/ops/cargo_new.rs#L741

JijoBose: I have addressed your code review pointers

**If you don't want a repository:**

```bash
topcoat new my-app --vcs none
```

Creates the project only — no `.git`, and no ignore file, since nothing would read it.

**If you want git (this is the default, so no flag needed):**

```bash
topcoat new my-app
```

Creates the project plus a git repository and a `.gitignore`.

**One catch:** the default skips the repository when the project lands inside an existing git or Mercurial checkout, so it doesn't nest one repo in another. If you're in that situation and want a repository anyway, ask for it:

```bash
topcoat new my-app --vcs git
```

Other systems work the same way — `--vcs hg`, `--vcs pijul`, `--vcs fossil` — and each gets the ignore file it actually reads.

ArjixWasTaken: Having a `.gitignore` even if the project is not tracked by git, is not a bad thing.
And the `.gitignore` could be considered part of the template, not part of the VCS.

I frequently find myself creating projects w/o a git repo, until I've written enough code to warrant version tracking.
And you can imagine how surprised I am when I see 800+ items added to git when I run `git add .`

JijoBose: Fair point, and it isn't harmful. I tied it to --vcs to match `cargo new`, which also writes no .gitignore for --vcs none or inside an existing repo, so the `git add .` surprise isn't something new here 😄

crates/topcoat-cli/src/new/template.rs
```diff
@@ -0,0 +1,522 @@
...
+                source.dependency(&["tailwind"], true),
+                format!(
+                    "\n[build-dependencies]\ntopcoat = {}\n",
+                    source.dependency(&["tailwind"], false)
+                ),
+            ),
+            Template::Minimal | Template::Runtime => (source.dependency(&[], true), String::new()),
+        };
+
+        format!(
```
atamakahere-git, line 154: Would be better to just use raw strings (`r#""#`), 
Ex:
```rs
fs::write(
        paths::home().join(".gitconfig"),
        r#"
        [init]
            defaultBranch = master
        "#,
    )
    .unwrap();
```
Ref: https://github.com/rust-lang/cargo/blob/0158e40d8638a7de292b7242b1533caaf48cbe5f/tests/testsuite/new.rs#L20-L27

atamakahere-git: Oh wait, this might be irrelevant if you are using {} inside format!(), But you can prefer using raw string where format is not required.

JijoBose: Raw strings and format! actually compose fine, the {{ / }} escaping is needed either way, so that's not a reason to avoid them. But your instinct lands in the right place: the template constants use no format! at all, and that's where most of it is, around 40 escaped quotes. Converting those.

JijoBose: Updated

atamakahere-git: [0001-chore-move-template-strign-to-files.patch](https://github.com/user-attachments/files/30495000/0001-chore-move-template-strign-to-files.patch)

Open for discussion, Having template "string" in files would make it more maintainable. Let's wait for a word from contributor/maintainer on this.

pikaju:
Sorry, I appreciate the effort but this isn't exactly how I envisioned it, I want to think about it more in the future

PR #194, opened 2026-07-24
feat: add static site export

carllerche:
## Summary

- Add the `topcoat export` CLI command for building static sites
- Support static asset files and generated values for dynamic route segments
- Add page-level and module-router APIs for static parameter generation
- Document static export usage and route generator behavior

## Testing

Not run (not requested)


carllerche:
@pikaju said this sucked and I should try again. #196 is attempt number 2.

PR #197, opened 2026-07-24
perf: template buffer preallocation, lazy header map initialization, and route param fast-path

harihar-nautiyal:
## Summary

1. `topcoat-view` & `topcoat-view-grammar`
**ViewParts Buffer Pre-allocation**: Updated `ViewWriter::into_token_stream` to pre-calculate node chunks and emit `ViewParts::with_capacity(#capacity)`. Added `with_capacity` to `ViewParts` to prevent `SmallVec`/heap reallocations during template assembly.
**Lazy HeaderMap Allocation**: Updated `Formatter` to lazily initialize `headers: Option<HeaderMap>` (`None` by default) instead of allocating `HeaderMap::new()` on every render pass, removing header heap allocations for standard view renders.

2. `topcoat-router`
**Path Parameter Fast-Path**: Added `if matched.params.is_empty()` check in `Router::handle` to return `RawPathParams::default()` directly for static endpoints (e.g., `/`, `/about`), avoiding key-value iteration overhead.

### Key Benefits
**Zero-Allocation Standard Renders**: Eliminates `HeaderMap` allocations when rendering views without custom headers.
**Improved Buffer Utilization**: `ViewParts` pre-allocates exact required capacity upfront.
**Faster Endpoint Dispatch**: Bypasses parameter slice parsing on static routes.
**100% Backward Compatible**: Pure internal performance optimizations without public API changes.

## Testing 
Tested without all features enabled
```
cargo test -p topcoat-view -p topcoat-view-grammar -p topcoat-router -p topcoat-asset -p topcoat-core
```

### DIFF
--- a/crates/topcoat-router/src/router.rs
+++ b/crates/topcoat-router/src/router.rs
@@ -72,7 +72,9 @@ impl Router {
         let (layers, terminal, path_params) =
             if let Ok(matched) = self.endpoints.at(parts.uri.path()) {
                 let endpoint = matched.value;
-                let path_params = {
+                let path_params = if matched.params.is_empty() {
+                    RawPathParams::default()
+                } else {
                     debug_assert_eq!(endpoint.path_params().len(), matched.params.len());
                     let keys = endpoint.path_params().iter().cloned();
                     let values = matched.params.iter().map(|(_, value)| value);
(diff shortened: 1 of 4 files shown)


pikaju:
I'm going to assume slop, most of these optimizations don't make a whole lot of sense. HeaderMap doesn't allocate either way this is just a useless extra branch

Issue #198, opened 2026-07-24
Creating signals dynamically (one per runtime collection item)


AmeinEskinder:
﻿The `signal` statement declares a single lexical binding, so there is no way to create one signal per element of a runtime collection:

```rust
// rows loaded from a database at request time
for row in &rows {
    signal value = row.default; // one binding, not one per row
}
```

Any UI rendered from runtime data (table rows, form builders, filter panels) runs into this: each item wants its own piece of client state, but the number of items is only known at request time. The fallback is a shard round-trip per interaction, which client-side signals exist to avoid.

Digging into `topcoat-runtime`, the underlying machinery does not appear to require compile-time signals at all — `SignalId` is generated at render time, the browser registry is keyed dynamically, and expression captures serialize `&Signal<T>` per render. Composing the public types directly:

```rust
let rows: Vec<(String, f64)> = load_rows();
let signals: Vec<Signal<f64>> = rows.iter().map(|(_, d)| Signal::new(*d)).collect();

view! {
    for s in &signals {
        (SignalDeclaration::new(s))            // declare each to the browser
    }
    for (i, (name, _)) in rows.iter().enumerate() {
        let sig = &signals[i];                 // view-`let` bridges collection -> capture
        <button @click=$(|_e| sig.set(sig.get() + 1.0))>"+"</button>
        " " $(sig.get())
    }
    let a = &signals[0];
    let b = &signals[1];
    <p>"Total: " $(a.get() * b.get())</p>      // cross-item dependency
}
```

This compiles and works in the browser on current `main`: per-item updates and the cross-item total re-render with zero network requests.

Is this pattern intended public API, or accidental surface? If it is intended, I'd be happy to send a PR documenting it (a "Dynamic signals" section in the runtime guide, plus an example and a regression test so the pattern doesn't silently break). If it is accidental, I'd be interested in the intended direction for this under the "(More) reactivity" roadmap item.

pikaju:
It's accidental, and the whole signal system will change so I'll close for now. We're reworking everything

PR #200, opened 2026-07-24
docs: document dynamic signal creation for runtime collections

AmeinEskinder:
﻿Documents the value-level path for creating signals dynamically — one per element of a runtime collection — raised in #198.

`Signal::new` and `SignalDeclaration` already compose into this; what was missing is documentation, an example, and tests so the pattern is stable API rather than accidental surface.

- **Guide**: a "Dynamic signals" section in the runtime guide, after bind attributes: creating signals programmatically, announcing them with `SignalDeclaration`, and bridging collection elements into expression captures with view-level `let`. Snippets run as doctests.
- **Example**: `examples/dynamic-signals`, an order form whose line items come from data — per-item steppers and a cross-item total, all client-side.
- **Tests**: `crates/topcoat-view/macro/tests/dynamic_signals.rs` pins the two behaviors the pattern relies on: each runtime-created signal renders a declaration with a distinct id, and capturing expressions hydrate those same ids (including one expression capturing two collection signals).

No behavior changes; `topcoat-view-macro` dev-dependencies gain the `runtime` feature for the new tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -703,6 +703,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "dynamic-signals"
+version = "0.4.0"
+dependencies = [
+ "tokio",
+ "topcoat",
+]
+
+[[package]]
 name = "either"
 version = "1.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


pikaju:
Thanks, but I don't think this makes a whole lot of sense, also we want to completely change how signals work so this wouldn't be useful for long

PR #201, opened 2026-07-24
fix: browser hang when a text expression renders an owned string

### LINKED ISSUE #192: Browser runtime hangs in an infinite loop when a text expression renders an owned `String`
## Summary

Any `$(...)` text expression whose value is an owned `String` surrogate (for example `$(string_signal.get())` or anything derived from `e.target.value` / `.to_owned()`) locks up the browser tab the first time it re-renders on the client. The main thread spins forever at 100% CPU and the page never responds again.

`&str` values are unaffected, and the initial server-side render is fine — the hang happens on the first client-side update (e.g. the first `signal.set(...)` that touches the expression).

## Reproduction

`topcoat = "0.4.0"` (`topcoat-runtime` 0.4.0), tested on Chrome 150 / macOS.

```rust
use topcoat::{
    Result,
    asset::{AssetBundle, RouterBuilderAssetExt},
    runtime::Event,
    view::{component, view},
};

#[tokio::main]
async fn main() {
    topcoat::start(
        Router::builder()
            .discover()
            .assets(AssetBundle::load().unwrap())
            .build(),
    )
    .await
    .unwrap();
}

#[page("/")]
async fn home() -> Result {
    view! {
        <!DOCTYPE html>
        <html>
            <head>
                topcoat::runtime::script()
            </head>
            <body>
                demo()
            </body>
        </html>
    }
}

#[component]
async fn demo() -> Result {
    view! {
        signal message = String::new();

        <button value="hello" @click=$(|e: Event| message.set(e.target.value))>
            "Set message"
        </button>
        <p>"Message: " $(message.get())</p>
    }
}
```

Steps:

1. `topcoat dev`, open the page.
2. Click "Set message".
3. The tab freezes permanently (infinite loop on the main thread; even `1+1` via DevTools `Runtime.evaluate` never returns).

Setting the signal to a constant (`message.set("hi".to_owned())`) hangs the same way, so it is not related to reading `e.target`.

## Root cause

`toText` in `browser/src/text.ts` unwraps ref-like values in a loop:

```ts
function toText(value: unknown): string {
	let current = value;
	while (isRefLike(current)) {
		current = current.deref();
	}
	...
}

function isRefLike(value: unknown): value is { deref: () => unknown } {
	return (
		value !== null &&
		typeof value === "object" &&
		typeof (value as { deref?: unknown }).deref === "function"
	);
}
```

but the owned-string surrogate in `browser/src/surrogate/string.ts` derefs to itself (mirroring Rust's `Deref<Target = StrSurrogate>`, where both sides share the same JS representation):

```ts
export class String extends Str {
	deref(): Str {
		return this;
	}
	...
}
```

`String` always stays ref-like, so the `while` loop never terminates. `Str` has no `deref`, which is why `&str` values render fine, and attribute binds are unaffected because they go through `toAttributeValue` instead of `toText`.

`$(string_signal.get())` always hits this because `Signal.get()` returns `this.read().deref().clone()`, i.e. an owned `String` instance.

## Possible fixes

- Make `String.deref()` return `new Str(this.v)` so the unwrap loop terminates, or
- In `toText`, check `isNodeViewParts` (i.e. `toNodeText`) *before* the deref-unwrap loop (both `Str` and `String` already implement `NodeViewParts`), or
- Bail out of the loop when `deref()` returns the same object.

## Workaround

Render the value server-side through a `#[shard]` (pass the signal as an argument) or bind it to an attribute (`:value=$(...)`) instead of interpolating it as text.

AmeinEskinder:
﻿Fixes #192.

`String.deref()` returned `this`, so `toText`'s ref-unwrapping loop (`while (isRefLike(current)) current = current.deref()`) never terminated: the first client-side re-render of a text expression producing an owned string spun the main thread forever. `$(string_signal.get())` always hit it, since `Signal.get()` returns an owned `String`.

The fix takes the first option from the issue's excellent diagnosis: `String.deref()` now returns `new Str(this.v)` — the borrowed form — mirroring the Rust surrogate, where `StringSurrogate` implements `Deref<Target = StrSurrogate>` rather than dereferencing to itself. Unwrap loops terminate structurally, and compiled `*expr` derefs still receive a `Str` with the full borrowed vocabulary.

- `src/surrogate/string.test.ts` pins the two properties the hang violated: deref yields the borrowed form (not `this`), and ref-unwrapping an owned string terminates.
- `dist/index.js` rebuilt (`pnpm build`).

Verified against the issue's reproduction: before the change the click freezes the tab (renderer pinned at 100% CPU); after it, `Message: hello` renders immediately and stays reactive.


pikaju (approved):
Thank you!

PR #203, opened 2026-07-25
docs: document creating signals dynamically for runtime collections

AmeinEskinder:
﻿Creating signals dynamically — one per element of a runtime collection — currently works but is undocumented and untested, so it is easy to break by accident.

The `signal` statement declares a single lexical binding, which fits state a component knows about while it is written. Markup rendered from data — table rows, form builders, filter panels — needs one signal *per item*, and the item count is only known at run time. `Signal::new` and `SignalDeclaration` already compose into exactly that: build one signal per item, render a declaration node for each, and bridge collection elements into expression captures with a view-level `let`.

This documents that path and pins it:

- **Guide**: a "Dynamic signals" section in the runtime guide, after bind attributes. Snippets run as doctests.
- **Example**: `examples/dynamic-signals` — an order form whose line items come from data, with per-item steppers and a cross-item total, all client-side.
- **Tests**: `crates/topcoat-view/macro/tests/dynamic_signals.rs` pins the two properties the pattern depends on — each runtime-created signal renders a declaration with a distinct id, and capturing expressions hydrate those same ids (including one expression capturing two collection signals).

No behavior changes. `topcoat-view-macro`'s dev-dependency on `topcoat` gains the `runtime` feature for the new tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -703,6 +703,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "dynamic-signals"
+version = "0.4.0"
+dependencies = [
+ "tokio",
+ "topcoat",
+]
+
+[[package]]
 name = "either"
 version = "1.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


pikaju:
See #200

PR #206, opened 2026-07-25
feat: add websocket shard support

seanaye:
This PR adds support for websocket shards via `#[shard(ws)]`. A websocket shard has a special fn signature which can reactively receive new messages from the client when a signal changes and must return an impl Stream<Item = Response>.

The existing shard combobox example has been updated to use this new feature.


pikaju:
I appreciate the effort but I don't think this is the direction we want to go in for now, the reactive system will be overhauled completely.

seanaye:
@pikaju understood, let me know if there are other things I can work on re: topcoat.

PR #214, opened 2026-07-27
feat(runtime): add utility methods for signals with primitive types

### LINKED ISSUE #175: add utility methods for signals with primitive types
* a `Signal<bool>` could have a method `.toggle()`
* a `Signal<f64>` could have `.increment()` and `.decrement()`
* a `Signal<String>` could have `.push_str(...)`
* etc

RAprogramm:
## Summary

Adds the convenience writes from #175 to the signal vocabulary: `toggle` on `Signal<bool>`, `increment` and `decrement` on `Signal<f64>`, and `push_str` on `Signal<String>`.

Each is an inherent method on the concrete `SignalSurrogate<T>`, so calling `toggle()` on a numeric signal is a compile error, and each is mirrored one-for-one in `WriteSignal` under the same name, since the expression compiler writes the Rust method identifier verbatim into the generated JavaScript. The browser side goes through the maverick updater form and constructs a new surrogate rather than mutating the stored one, which is what keeps identity-based change detection notifying subscribers. Server-side evaluation panics with the message `set` already uses, now shared by a small helper that `set` also calls.

Deliberately out of scope: no `update`/`with` closure primitive, no assignment operators in the expression grammar, no `Ref::deref_mut` write-through, no `increment_by`, no integer or container signals, and no change to the README example. The general mapping-closure primitive looks like the better long-term answer and is worth its own issue; this patch is the sugar the issue asks for, not a replacement for it.

The reasoning behind the shape, the three designs that were rejected and the constraints that decide it are written up here: https://raprogramm.web.app/articles/signal-mutation-api-under-dual-compilation

Closes #175

## Testing

```
cargo +nightly fmt --all
cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
cargo test --workspace --all-features
RUSTDOCFLAGS="--cfg docsrs -Dwarnings" cargo +nightly doc --workspace --all-features --no-deps --locked
cargo outdated --workspace --root-deps-only
cargo +nightly udeps --package topcoat-runtime-macro --all-targets --all-features --locked
```

All clean. In `crates/topcoat-runtime/browser`: `yarn install --frozen-lockfile`, `yarn build`, `yarn test` (7 passing, 4 of them new), and the regenerated `dist/index.js` is committed.

New tests: one `#[should_panic]` per method for the server-side write, three compile-through tests in `crates/topcoat-runtime/macro/tests/expr.rs` asserting the method name reaches the emitted JavaScript, and four vitest cases covering the value changes and that a write notifies subscribers exactly once.

Manually verified against `topcoat dev` in `examples/runtime`: `/counter` renders `__external0.increment()` and `__external0.decrement()`, `/show` renders `__external0.toggle()`.

Two manifest notes, both in `crates/topcoat-runtime/macro/Cargo.toml`: the compile-through test needs `tokio` as a dev-dependency, because a `view!` body with elements expands to code that awaits and the repository's idiom for that is `#[tokio::test]`; and with a test now using the `topcoat` dev-dependency, the `cargo-udeps` ignore entry is no longer needed and was removed.

## Questions

1. `increment`/`decrement` are fixed at a step of one. Is `increment_by(f64)` wanted, or is `set(count.get() + n)` the intended spelling?
2. Should the "etc" in the issue be taken up now (`Option::take`, `Vec::push`), or left to separate issues? This assumes the latter.
3. Is a defensive `instanceof` check wanted in the TypeScript methods, given the Rust type check already makes a wrong call unrepresentable?
4. `push_str` takes `&StrSurrogate` because a string literal lowers to exactly that, and since arguments are emitted by value it cannot accept the result of `get()`. This affects every existing borrowing method equally; worth its own issue?


pikaju (approved):
Thank you!

PR #215, opened 2026-07-27
feat(router): validate websocket handshake keys

aodihis:
Hi, I've been experimenting with and testing Topcoat for a while. I noticed that, during the WebSocket handshake, Topcoat doesn't validate whether Sec-WebSocket-Key is a valid Base64-encoded 16-byte value.

This PR adds a small validation check, along with test cases for valid and invalid keys. The implementation is modeled after the validation used by [ws](https://github.com/websockets/ws/blob/master/lib/websocket-server.js#L263).

If this doesn't align with the project's preferred approach, I'm happy to revise the implementation or close the PR.


pikaju (approved):
Thank you!

PR #225, opened 2026-07-28
docs(router): add design for page URL generation

carllerche:
[Rendered](https://github.com/tokio-rs/topcoat/blob/url-generation-design/design/url_generation.md)

## Summary

Design for generating a page's URL from its path parameters, so links name the handler they point at instead of a hardcoded string.

`href!(post, PostId(42))` expands to `Href::new(post, (PostId(42),))`, passing the marker struct `#[page]` already generates and the parameters as a tuple. It returns a lazy `Href` that resolves against the route table and base URL when rendered, mirroring how `Asset` resolves through the app context. A page with no parameters renders as its own URL, so `href=(about)` is a link. `href!` is the only macro: an `Href` renders relative in a view and absolute in mail, and `.absolute()`/`.relative()` override that per link.

Parameters are checked when the URL is built rather than at compile time. A macro cannot see an ancestor module's `#[path_param]`, and a module-derived path is not assembled until `Router::build()`, so a compile-time check would hold in some modules and not others. Failures carry the source location of the `href!` that produced them, so the error names the link rather than the page rendering it.

**This is a design PR. The implementation follows once the direction is approved.**

The doc closes with an "Other ideas" section for the parts that are not settled: taking the query string and fragment as macro arguments, test assertions over links, and the `module_router!` change that would make every page compile-time checked.

## Testing

No code, so no test suite. The mechanisms the design leans on were verified against rustc with standalone programs:

- an explicit item shadows a glob import
- a `#[used]` linker-section static inside a block registers before `main`, including from functions that never run
- that static cannot live in a generic fn (E0401), and cannot infer its type from a sibling expression
- `#[track_caller]` reports the `href!` call site through `macro_rules!`
- an `include!`-based module tree accumulates the parameter chain and preserves `module_path!()`

The third result is the load-bearing one: it rules out collecting call-site parameter *types* at link time, which is why there is no `Router::build()` validation pass and why checking happens where the types exist.

### DIFF
--- a/design/url_generation.md
+++ b/design/url_generation.md
@@ -255,14 +255,6 @@ Reading would special-case `String` the way it special-cases `str` today, so ser
 
 One trait impl on the marker. The risk is two ways to write a link, with the shorter one breaking the moment the page gains a parameter.
 
-**Read a parameter list in attribute position.** `view!` could treat a parenthesized list in `href` or `action` position as a page plus its parameters.
-
-```rust
-<a href=(post, PostId(42))>"Post"</a>
-```
-
-Shortest form, no import. It also gives `(a, b)` a meaning in two attributes that it has nowhere else in `view!`, where parenthesized values are plain Rust.
-
 **Choose the attribute from the element.** A `page` attribute could expand to `href` on an anchor and `action` on a form.
 
 ```rust
@@ -270,7 +262,7 @@ Shortest form, no import. It also gives `(a, b)` a meaning in two attributes tha
 <form page=(create) method="post">
 ```
 
-Removes the last duplication, but invents a non-HTML attribute and hides which one it emits.
+Overloading `href` instead is not available. `AttributeValueViewParts` is implemented for tuples, so a parenthesized list in attribute position already means "render these in order", and `href=(post, PostId(42))` has a meaning today. A new attribute name sidesteps that, at the cost of inventing one that is not HTML and hiding which attribute it produces.
 
 **Check parameters at compile time.** Giving `module_router!` the route tree changes what the macro knows. It can pull in each module body itself, thread a module's parameters down to its children, and generate a typed `href` per page.
 
--- a/design/url_generation.md
+++ b/design/url_generation.md
@@ -21,18 +21,19 @@ href!(post, PostId(42))                       // "/posts/42"
 href!(user, OrganizationId(9), UserId(41))    // "/organizations/9/users/41"
 ```
 
-Path parameters are the [`#[path_param]`](../crates/topcoat-router/macro/docs/path_param.md) newtypes the page already reads from the request, so one type serves both directions.
+Path parameters are the [`path_param!`](../crates/topcoat-router/macro/docs/path_param.md) newtypes the page already reads from the request, so one type serves both directions.
 
 ```rust
 // src/app/posts/post_id.rs
-#[path_param(error = not_found)]
-pub struct PostId(pub u64);
+path_param!(post_id: u64, error = not_found);
 
 #[page]
 async fn post(cx: &Cx) -> Result {
     view! { <h1>"Post " (path_param::<PostId>(cx)?)</h1> }
 }
 ```
+
+`path_param!` declares the parameter by its URL name and generates the type, `PostId` here. It replaces today's `#[path_param]` attribute, which cannot declare the generic that a string parameter needs to be constructible. This document assumes that change; its design lands next.
 
 ```rust
 // src/app/posts.rs
@@ -110,20 +111,20 @@ href!(document, DocPath("guides/getting-started"))    // "/docs/guides/getting-s
 
 An empty value leaves nothing between the slashes, and the `/posts/` it produces no longer matches the route it was built from, so it is rejected when the URL is built.
 
-A parameter the page reads as a string has to be constructible from a string. `#[path_param]` declares those over `str` today, which is unsized and so cannot be constructed at all, so an unparsed parameter takes its inner type as a type parameter instead.
+A parameter declared without a type is unparsed: the page reads its segment as a string. Reading alone could hold that string as an unsized `str`, but a link has to construct the value, so the generated type is generic over anything that borrows as one.
 
 ```rust
-pub struct Slug<T: AsRef<str>>(pub T);
+path_param!(slug);      // pub struct Slug<T: AsRef<str>>(pub T);
 
 href!(show, Slug("my-first-post"))    // &str
 href!(show, Slug(post.slug))          // String
 ```
 
-Reading is unchanged: `path_param::<Slug>(cx)` still borrows the decoded segment out of the request and allocates nothing. Declaring a generic tuple struct through an attribute macro reads poorly, so the declaration form belongs with [`path_param`](../crates/topcoat-router/macro/docs/path_param.md) rather than here.
+Reading is unaffected: `path_param::<Slug>(cx)` borrows the decoded segment out of the request, allocating nothing. The allocation, if there is one, moves to the link, where the value usually came out of a `String` already.
 
 Group segments never appear in a served URL and take no value, so a page under `app::_marketing::pricing` is reached with `href!(pricing)`.
 
-Linking to a page from outside its own module needs its parameter type, so declare the type and its field `pub`.
+Linking to a page from outside its own module needs its parameter type, so `path_param!` generates a public type with a public field.
 
 # When mistakes are caught
 
--- a/design/url_generation.md
+++ b/design/url_generation.md
@@ -50,64 +50,76 @@ async fn index() -> Result {
 }
 ```
 
-`href!` rewrites to a method on the page, passing the parameters as a tuple.
+`href!` expands to `Href::new`, passing the page marker that `#[page]` already generates and the parameters as a tuple.
 
 ```rust
 href!(user, OrganizationId(9), UserId(41))
-user.href((OrganizationId(9), UserId(41)))
+Href::new(user, (OrganizationId(9), UserId(41)))
 ```
 
-The method takes exactly one argument because Rust has no variadic functions, and one argument holding any number of parameters means a tuple. Zero and one read worst.
+The constructor takes a single argument for the parameters because Rust has no variadic functions, so any number of them means a tuple. Zero and one read worst, and zero is what most pages take.
 
 ```rust
-about.href(())
-post.href((PostId(42),))
+Href::new(about, ())
+Href::new(post, (PostId(42),))
 ```
 
-Most pages have no path parameters, so `about.href(())` is the shape a reader meets most often. `href!` spreads the parameters into that tuple, keeping it out of every link.
+`href!` spreads its arguments into that tuple, keeping it out of every link.
 
-The tuple stays the real API. It gives every page one calling convention, and leaves a method for helpers that build URLs generically. Everything below applies to both forms.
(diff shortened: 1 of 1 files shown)


design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+
+```rust
+<a href=(href!(post, PostId(42)))>"Post"</a>
+```
+
+Elsewhere it takes a context.
+
+```rust
+href!(post, PostId(42)).resolve(cx)      // "/posts/42"
+```
```
pikaju, line 76: I think this is great. `Href` should implement `AttributeValueViewParts` and `NodeViewParts`. It should also emit each URL part as one `ViewPart` to leverage Topcoat's lazy rendering, there is prior work for this with `class!` and `Class<T>`.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+    }
+}
+```
+
+`href!` rewrites to a method on the page, passing the parameters as a tuple.
+
+```rust
+href!(user, OrganizationId(9), UserId(41))
+user.href((OrganizationId(9), UserId(41)))
+```
```
pikaju: I think if we have the macro we don't need the method on pages anymore, it just adds code gen noise. Instead, maybe the `Href` struct (see below) should be called with `Href::new(page, (param1, param2))`. I would argue that keeps the functionality nicely encapsulated

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+
+```rust
+href!(index).query(&[("page", 2)])
+```
+
+`fragment` appends a `#` fragment.
+
+```rust
+href!(post, PostId(42)).fragment("comments")      // "/posts/42#comments"
+```
```
pikaju, line 179: This looks fine to me, although we _could_ also just put queries into the href itself, e.g. `href!(my_page, PostsQuery { page: Some(2) })`. Although this wouldn't work with `.fragment` so maybe `.query` is fine.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+
+`href!` produces a root-relative URL, which is what a link inside the site needs. Content that leaves the site, such as mail, feeds, and sitemaps, needs the absolute form. `url!` takes the same arguments and resolves against the [base URL](../crates/topcoat/docs/context.md) registered on the router.
+
+```rust
+let router = Router::builder().base_url("https://example.com").build();
+
+url!(post, PostId(42))        // "https://example.com/posts/42"
+```
+
+An application mounted under a path prefix registers it as part of that base URL, as in `https://example.com/app`. The proxy strips the prefix before the router matches, so the router only ever sees `/posts/42` while the browser needs `/app/posts/42`. Relative URLs carry the prefix for that reason, so `href!` reads the base URL too.
```
pikaju: I'm not a big fan of this because `href` doesn't mean "relative", and this adds yet another macro. Wouldn't it be nice to just do `href(page).absolute()`? I think that's pretty hot. More characters, yes, but very sexy. And keeps all the responsibility on the `Href` struct and macro.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+let url = url!(post, PostId(42)).resolve(&cx);
+```
+
+Tests use the same method. A module-derived URL resolves through the route table, so a context built by hand covers only pages with an explicit path.
+
+```rust
+let cx = app::router().cx();
+assert_eq!(href!(post, PostId(42)).resolve(&cx), "/posts/42");
+```
+
```
pikaju, line 239: Good, I like `.resolve(cx)`. `view!` doesn't need it because it calls `into_view_parts` instead.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+`#[route]` handlers work the same way, which covers form actions and fetch targets.
+
+```rust
+#[route(POST "/api/posts/{post_id}/publish")]
+async fn publish(cx: &Cx) -> Result<&'static str> {
+    Ok("published")
+}
+
+href!(publish, PostId(42))        // "/api/posts/42/publish"
+```
```
pikaju, line 251: Yes, great.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+
+`query` and `fragment` extend an `Href` before it resolves, and the redirect constructors take one as is. Comments below show what a URL resolves to.
+
+# Parameter values
+
+Each value is percent-encoded into its segment, so a title or an address containing `/`, `?`, or `#` stays inside the segment it belongs to.
+
+```rust
+href!(show, Slug::new("hello/world"))     // "/posts/hello%2Fworld"
```
pikaju: I think we want to avoid `::new` on the path parameter structs because it's so noisy in this context. More thoughts below.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+```rust
+<a href=(about)>"About"</a>
+```
+
+One trait impl on the marker. The risk is two ways to write a link, with the shorter one breaking the moment the page gains a parameter.
+
+**Choose the attribute from the element.** A `page` attribute could expand to `href` on an anchor and `action` on a form.
+
+```rust
+<a page=(post, PostId(42))>"Post"</a>
```
pikaju: This isn't valid syntax at the moment, it should be `href=((post, PostId(42)))`

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+Reading would special-case `String` the way it special-cases `str` today, so serving a request still allocates nothing. The allocation moves to URL building, where the value usually came out of a `String` anyway. `Cow<'static, str>` avoids it for literals at the cost of a noisier type.
+
+`String` then reads identically to `str` and does more, so the question is whether `str` stays at all.
+
+**Render a page marker directly.** A page with no parameters could render as its own URL.
+
+```rust
+<a href=(about)>"About"</a>
```
pikaju: Yes I was planning on doing this for a while. I think the implementation of this should use `Href::new` internally to be consistent.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+**Choose the attribute from the element.** A `page` attribute could expand to `href` on an anchor and `action` on a form.
+
+```rust
+<a page=(post, PostId(42))>"Post"</a>
+<form page=(create) method="post">
+```
+
+Overloading `href` instead is not available. `AttributeValueViewParts` is implemented for tuples, so a parenthesized list in attribute position already means "render these in order", and `href=(post, PostId(42))` has a meaning today. A new attribute name sidesteps that, at the cost of inventing one that is not HTML and hiding which attribute it produces.
+
+**Check parameters at compile time.** Giving `module_router!` the route tree changes what the macro knows. It can pull in each module body itself, thread a module's parameters down to its children, and generate a typed `href` per page.
```
pikaju: I like the idea but I think we should not do this for the first iteration, let's just stick to dynamic checks for now.

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+One trait impl on the marker. The risk is two ways to write a link, with the shorter one breaking the moment the page gains a parameter.
+
+**Choose the attribute from the element.** A `page` attribute could expand to `href` on an anchor and `action` on a form.
+
+```rust
+<a page=(post, PostId(42))>"Post"</a>
+<form page=(create) method="post">
+```
+
+Overloading `href` instead is not available. `AttributeValueViewParts` is implemented for tuples, so a parenthesized list in attribute position already means "render these in order", and `href=(post, PostId(42))` has a meaning today. A new attribute name sidesteps that, at the cost of inventing one that is not HTML and hiding which attribute it produces.
```
pikaju: Ah, no I hate this

pikaju: We can just stick to `href=(href!(page, ...))` for hrefs with params I think that's fine. Parameterless pages could still work with `href=(page)`

design/url_generation.md
```diff
@@ -0,0 +1,283 @@
...
+#[path_param]
+pub struct Slug(pub String);
+
+path_param::<Slug>(cx)                  // still &str, still no allocation
+href!(show, Slug(post.slug.clone()))
+```
+
+Reading would special-case `String` the way it special-cases `str` today, so serving a request still allocates nothing. The allocation moves to URL building, where the value usually came out of a `String` anyway. `Cow<'static, str>` avoids it for literals at the cost of a noisier type.
+
+`String` then reads identically to `str` and does more, so the question is whether `str` stays at all.
```
pikaju: After thinking about it I think I might agree with @carllerche that we could revamp the `path_param` macro to look like this:

```rust
path_param!(my_param); // unparsed &str case, generates `struct MyParam<T: AsRef<str>>(pub T)`
path_param!(my_param: MyType); // parsed case, generates `struct MyParam(pub MyType)`

// and maybe
path_param!(my_param: MyType, error = not_found); // handle errors like before
```

The `path_param::<T>(cx)` function should work the same way and return `&str` (no result) for the unparsed case but now you can construct the struct with an owned String `MyParam(owned)` to, for example, return it from a function.

design/url_generation.md
pikaju: Also just to clarify the result type should probably be another `Href`

pikaju (changes requested):

design/url_generation.md
```diff
@@ -0,0 +1,291 @@
...
+post.href((PostId(42),))
+```
+
+Most pages have no path parameters, so `about.href(())` is the shape a reader meets most often. `href!` spreads the parameters into that tuple, keeping it out of every link.
+
+The tuple stays the real API. It gives every page one calling convention, and leaves a method for helpers that build URLs generically. Everything below applies to both forms.
+
+# What href! returns
+
+An `Href`, not a `String`. Resolving one needs the route table and the base URL, both of which live on the request context, so an `Href` describes a URL and builds it later. That is also why it cannot implement `Display`.
```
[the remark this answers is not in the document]
carllerche: The default could be `localhost:3000` or whatever port the dev server is listening on.

design/url_generation.md
```diff
@@ -0,0 +1,291 @@
...
+href!(document, DocPath::new("guides/getting-started"))    // "/docs/guides/getting-started"
+```
+
+A parameter declared over `str` is unsized and cannot be built with the tuple-struct constructor, so `#[path_param]` gives those types a `new` constructor that borrows the string.
+
+```rust
+#[path_param]
+pub struct Slug(str);
+
+href!(show, Slug::new("my-first-post"))
```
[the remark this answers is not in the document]
carllerche: Validating them when the URL is created works.

design/url_generation.md
carllerche: We also could do `href!(post, PostId(42), fragment: "comments");` (same with `query: `.

```rust
href!(index, query: PostsQuery { page: 2, q: "rust" });
href!(index, query: { "page": 2 });
href!(post, PostId(42), fragment: "comments");
```

For this to work, #[query_params] would need to generate a builder like Toasty does and `query: PostsQuery { ... })`. expands to something similar to the `create!` macro.

design/url_generation.md
carllerche: I think that is fine, only one macro: `href!`. Default render is relative in html views, absolute in contexts like email. `.absolute()` and `.relative()` toggles the rendering.

design/url_generation.md
carllerche: We could also add special assertions for tests if we need to that handle resolution and possibly give other nice-to-have features for asserting URL links

design/url_generation.md
carllerche: Yes, but it *could* be valid syntax :)

pikaju:
implemented in #350

PR #230, opened 2026-07-28
fix(runtime): reject invalid procedure signatures at parse time

k5602:
## Summary

`#[procedure]` used to accept any function signature at parse time and then blow up later. A `self` receiver or a missing return type hit `unreachable!` in `Procedure::to_tokens`, so users got `error: proc macro panicked` instead of a real diagnostic. A missing `async` made it through as well and failed deep inside the generated code with `` `Result<T>` is not a future ``.

This moves the checks into `ProcedureItem::parse`, the same place `ShardItem` already validates its signature: procedures must be async, must have a return type, and must not take a `self` receiver. Errors are spanned to the offending tokens and surface through the existing `to_compile_error()` path. The two `unreachable!`s in `to_tokens` stay as internal invariants, relabeled `validated by ProcedureItem` to match `shard.rs`.

Destructured argument patterns like `async fn f((x, y): (f64, f64))` stay valid on purpose: the generated call only reads argument types, and there is a test pinning that.

## Testing

- `cargo test -p topcoat-runtime-grammar` (5 new tests: three rejections, two acceptances)
- `cargo fmt --all`, `cargo topcoat fmt`
- nightly `cargo doc --workspace --all-features --no-deps --locked` with `-Dwarnings`

Any feedback is welcome and thanks.

### DIFF
--- a/crates/topcoat-runtime/grammar/src/procedure.rs
+++ b/crates/topcoat-runtime/grammar/src/procedure.rs
@@ -3,6 +3,7 @@ use quote::{ToTokens, format_ident, quote};
 use syn::{
     FnArg, ItemFn, Pat, PatIdent, PatType, ReturnType,
     parse::{Parse, ParseStream},
+    spanned::Spanned,
 };
 use topcoat_core_grammar::paths::{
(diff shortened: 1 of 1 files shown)


pikaju (approved):
Thank you!

PR #231, opened 2026-07-28
docs(router): add design for the path_param! macro

carllerche:
[Rendered](https://github.com/tokio-rs/topcoat/blob/path-param-macro-change/design/path_param.md)

## Summary

Design for replacing the `#[path_param]` attribute with `path_param!`, the change proposed in the review of #225.

A declaration names the parameter the way the URL does and the macro generates the type: `path_param!(pub post_id: u64, error = not_found)` gives `pub struct PostId(pub u64)`, and `path_param!(pub slug)` gives `pub struct Slug<T: AsRef<str> = String>(pub T)`. The attribute went the other way, snake-casing the struct name into a URL name, and it could not give the unparsed form the generic it needs. `struct Slug(str)` reads a request fine but its field is unsized, so no link can build one, which is what the URL generation design needs.

Reading is unchanged. `path_param::<T>(cx)` still returns `Result<&T, ..>` for a parsed parameter and `&str` for an unparsed one, memoized per request. The `String` default on the type parameter covers the two positions with nothing to infer from: `path_param::<Slug>(cx)` and a signature that returns `Slug`.

Two points go past what the review thread settled, and are called out in the document:

- **Typed catch-alls**, `path_param!(pub *doc_path)`. Today a catch-all is `segment!(kind = CatchAll, rename = "path")` with no typed accessor, so `href!(document, DocPath(..))` in #225 has no type to pass.
- **The default type parameter**. #225 writes `Slug<T: AsRef<str>>` with no default; without one, reading and signatures both need a type argument. That snippet should pick up the default.

The document ends with migration notes. `#[path_param]` is removed rather than deprecated.

**This is a design PR. The implementation follows once the direction is approved.**

## Testing

No code, so no test suite. The claims about current behavior were read out of the implementation: the emitted `segment!(kind = Param, rename = ...)` and the missing-parameter panic in `crates/topcoat-router/grammar/src/path_param.rs`, the `duplicate segment specifier` panic in `crates/topcoat-router/src/module/segment.rs`, and the per-request decoding in `RawPathParams::from_pairs`.

That last one turns up a correction for #225: "so serving a request allocates nothing" is not true of unparsed parameters. `from_pairs` allocates a `Box<str>` per captured value before the handler runs; reading is what adds no allocation.

## Disclaimer

Written by an AI agent from the design document in #225 and its review comments, then reviewed against the router implementation. Reviewed and directed by @carllerche.

### DIFF
--- a/design/path_param.md
+++ b/design/path_param.md
@@ -20,14 +20,14 @@ One declaration serves both directions. The page reads a `PostId` out of the req
 A declaration is a visibility, the parameter name, and the type the segment parses into.
 
 ```rust
-path_param!(post_id: u64);                      // struct PostId(pub u64);
+path_param!(post_id: u64);                      // struct PostId(u64);
 path_param!(post_id: u64, error = not_found);   // the same type; a bad parse answers 404
 path_param!(pub slug);                          // pub struct Slug<T: AsRef<str> = String>(pub T);
 ```
 
-The name is spelled the way it appears in the URL, and the type is that name in Pascal case, so `post_id` matches `{post_id}` and generates `PostId`. The type and its field take the declared visibility.
+The name is spelled the way it appears in the URL, and the type is that name in Pascal case, so `post_id` matches `{post_id}` and generates `PostId`. The type and its field take the declared visibility, and the generic appears only on the unparsed forms, which do not fix the type of the value they hold.
 
-Most declarations need none. A parameter is read by the module that declares it and by its descendants, which reach a private type through ordinary Rust visibility, so the pages under `posts/post_id.rs` read `PostId` without it being `pub`. Linking is what needs `pub`, since the module holding the link is usually a parent or a sibling.
+A descendant module reads a private type through ordinary Rust visibility, so the pages under `posts/post_id.rs` read `PostId` without `pub`. A link from a parent or a sibling module needs it.
 
 The attribute went the other way, snake-casing the name of the struct you wrote into a URL name. Naming the parameter first puts the URL in the declaration and leaves the type to the macro, which is what the unparsed form needs: an attribute cannot add a generic without handing back a different struct than the one you wrote.
 
@@ -69,6 +69,7 @@ use topcoat::router::path_param;
 A parsed parameter returns a `Result` holding a reference to the parsed value, parsed at most once per request. The type has to implement `FromStr`, and the parse result has to be `Send + Sync + 'static` so it can be memoized.
 
 ```rust
+// with `error = not_found` on the declaration
 let post_id: &u64 = path_param::<PostId>(cx)?;
 ```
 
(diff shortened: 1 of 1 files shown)


design/path_param.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+`struct Slug(str)` reads a request fine, but its field is unsized, so no link can build one. That is what the unparsed form is for.
+
+# Catch-all parameters
+
+A `*` declares a parameter that matches the rest of the path. Its value is the remaining segments with the separators between them, so it is the one parameter that spans more than one segment.
+
+```rust
+path_param!(pub *doc_path);             // matches {*doc_path}, read as &str
```
pikaju: Maybe this should read as `&topcoat::router::Path` instead?

design/path_param.md
```diff
@@ -0,0 +1,152 @@
+`path_param!` declares a path parameter by the name it has in the URL and generates the type used to read it. It replaces the `#[path_param]` attribute.
+
+```rust
+// src/app/posts/post_id.rs
+use topcoat::{Result, context::Cx, router::{page, path_param}, view::view};
+
+path_param!(pub post_id: u64, error = not_found);
```
pikaju: Having `pub` as an option is good, but I think the default recommended case should be to have it non-pub, as, generally, only child modules need to have access to a path parameter of the parent

design/path_param.md
```diff
@@ -0,0 +1,152 @@
...
+An unparsed parameter has nothing to fail at, so `error` on one is a compile error.
+
+# Building a value
+
+The declaration produces a type you can construct, which is what a link needs.
+
+```rust
+use crate::app::posts::post_id::{PostId, post};
+
+href!(post, PostId(42))       // "/posts/42"
```
pikaju, line 105: `href!` not part of this design proposal though

pikaju (changes requested):

design/path_param.md
pikaju: Actually no, I think for catch-all parameters we want to use &[str] for the unparsed case and Vec<T> for parsed case (which can be returned by `path_param::<MyParam>(cx)` as `&[T]`

pikaju: Instead of `Vec<T>` the user can probably specify any collection that has `FromIterator<T>` so we parse each URL segment with FromStr and put it into the collection

design/path_param.md
carllerche: well yeah, but it is referencing the other.

design/path_param.md
```diff
@@ -0,0 +1,190 @@
...
+A declaration emits `segment!(kind = CatchAll, rename = "doc_path")`, which puts three rules on where it can appear:
+
+- One per module, like any other parameter. A second declaration panics with `duplicate segment specifier`.
+- Last served segment. A route with anything after the catch-all fails in `Router::build`, which panics with `failed to register route` carrying matchit's `InvalidCatchAll`.
+- One segment at least, so `/docs` does not reach a page under `{*doc_path}`.
+
+Reading yields the segments.
+
+```rust
+let doc_path: PathSegments<'_> = path_param::<DocPath>(cx);   // "guides", "getting-started"
```
pikaju, line 144: `PathSegments` is overloaded with the `PathSegment` we already have (which means something slightly different) and so I think `path_param::<DocPath>(cx)` should just return `impl Iterator<Item = &'cx str>`

pikaju:
PR is merged

PR #243, opened 2026-07-29
fix(asset)!: write the bundle next to the executable it was scanned from

### LINKED ISSUE #212: asset! declarations in the final binary crate are stripped by the MSVC linker on Windows
# `asset!` declarations in the final binary crate are stripped by the MSVC linker on Windows

  ## Summary

  On `x86_64-pc-windows-msvc`, any asset declared with `asset!` (or a macro expanding to it,
  such as `tailwind::stylesheet!()`) **from the binary crate itself** never reaches the bundler.
  The `#[used] static ENCODED_ASSET` is dropped at link time, so `Bundler::bundle` never sees the
  declaration, the asset is missing from `manifest.toml`, and the application panics at runtime as
  soon as a view tries to resolve it.

  Assets declared inside library crates (e.g. the `topcoat-runtime` browser script) are unaffected,
  which makes the failure look partial: some assets are served, the app's own ones are not.

  Linux and macOS are not affected.

  ## Reproduction

  ```
  cargo run -p tailwind
  ```

  (`examples/tailwind`, on Windows with the MSVC toolchain)

  ## Actual behaviour

  The page is served unstyled — the `Tailwind is not working` warning is visible — and requesting it
  panics:

  ```
  thread 'tokio-rt-worker' (32604) panicked at crates\topcoat-asset\src\config.rs:138:13:
  failed to resolve asset Asset(11515530482864368259) in the asset catalog
  ```

  The generated `target/assets/manifest.toml` contains only the runtime script, not the Tailwind
  stylesheet.

  ## Expected behaviour

  Same as on Linux: the stylesheet is bundled, `manifest.toml` lists both assets, and the page renders
  styled.

  ## Environment

  - Windows 11 Pro 
  - `rustc 1.97.1 (8bab26f4f 2026-07-14)`, host `x86_64-pc-windows-msvc`, LLVM 22.1.6
  - `cargo 1.97.1 (c980f4866 2026-06-30)`
  - topcoat `main` @ `ca09033`
  - Not reproducible on Linux

AmeinEskinder:
Fixes #212.

## What is still broken after #217

#217 did fix the MSVC linker problem. Byte-scanning both `tailwind.exe` builds for the `TOPCOAT_ASSET` marker on `x86_64-pc-windows-msvc` shows the declaration surviving in the release binary:

```
target\debug\tailwind.exe    id=5404681112881914444    path=...\target\debug\build\tailwind-7f722d3359c5bb6b\out/tailwind.css
target\release\tailwind.exe  id=12480077591628209191   path=...\target\release\build\tailwind-b97e50b9b147424f\out/tailwind.css
```

Nothing is stripped. But the two IDs differ, and that is the remaining bug. It is not Windows-specific.

`tailwind::stylesheet!()` expands to `asset!(concat!(env!("OUT_DIR"), "/tailwind.css"))`, and `AssetId::new` hashes the path string. `OUT_DIR` contains the profile, so the ID is profile-scoped.

`topcoat asset bundle` defaults to the `dev` profile, and every profile wrote its bundle to the same `<target>/assets`. So the reporter's `topcoat asset bundle` + `cargo build --release` leaves a dev-keyed manifest that the release binary cannot resolve:

```
thread 'tokio-rt-worker' panicked at crates\topcoat-asset\src\config.rs:139:13:
failed to resolve asset Asset { id: AssetId(12480077591628209191) } in the asset catalog
```

The two profiles also clobber each other's bundle, so they cannot coexist in one workspace.

## The change

Bundle into an `assets` directory next to the executable that was scanned, rather than one shared `<target>/assets`. `AssetBundle::load` already probes `<exe_dir>/assets` first, so no runtime change is needed: each profile finds the bundle built from it.

- `topcoat-cli/src/asset/bundle.rs`: `run_bundle` takes the scanned executable's path and defaults the output to `<exe_dir>/assets`. `--out` is unchanged.
- `topcoat-cli/src/dev/build.rs`: pass the executable through.
- `topcoat-cli/src/asset/clean.rs`: with no single fixed location left, remove every bundle under the target directory (`<target>/assets`, `<target>/*/assets`, `<target>/*/*/assets`), keyed on the directory actually holding a manifest so an unrelated `assets` directory is left alone. The `<target>/assets` entry also cleans up bundles left by earlier versions.
- `topcoat-asset/src/bundle.rs`: `AssetBundle::load` doc updated. No code change.
- `topcoat/docs/asset.md`: new default path, plus the `--release`/`--profile` flags (previously undocumented) and why the profile matters.

A mismatched bundle now fails at startup naming the directory it looked in, instead of panicking on the first request that renders an asset:

```
no asset bundle found near D:\Dev\topcoat\target\release:
tried D:\Dev\topcoat\target\release\assets, D:\Dev\topcoat\target\assets
```

## Verified

On `x86_64-pc-windows-msvc`, rustc 1.96.0, with `examples/tailwind`:

| Flow | Before | After |
|---|---|---|
| `asset bundle` + run release binary | request panics on an opaque asset ID | startup error naming the missing directory |
| `asset bundle --release` + run release binary | 200, only if nothing had clobbered `target/assets` | 200, `href="/_topcoat/assets/tailwind-a261247bbd892b1f.css"` |
| dev and release bundles at once | mutually clobbering | coexist, both 200 |
| `asset clean` | removed `<target>/assets` only | removes all bundles, including the legacy one |
| `topcoat dev` | — | 200, bundles into `target/debug/assets` |

`cargo fmt --check`, `cargo clippy` on both crates, and `cargo test -p topcoat-asset -p topcoat-cli` are clean.

## Not done here

Making the Tailwind ID profile-independent. The ID is defined as a hash of the declaring crate, source file, and path; special-casing generated paths cannot distinguish two different generated files, and it would hide the same mismatch for other assets.

`Debug for Asset` prints only the ID, which is why the original panic was hard to read. Decoding the embedded declaration so it also names the crate and path would help every stale-bundle case, but it is a separate change.

## Breaking change

The default bundle directory moves from `<cargo-target>/assets` to `<cargo-target>/<profile>/assets`. Anything copying or loading `target/assets` by hand needs updating. `AssetBundle::load` users are unaffected.

### DIFF
--- a/crates/topcoat-asset/src/bundle.rs
+++ b/crates/topcoat-asset/src/bundle.rs
@@ -1,5 +1,4 @@
 use std::{
-    ffi::OsStr,
     io,
     path::{Path, PathBuf},
 };
@@ -18,34 +17,24 @@ pub struct AssetBundle {
 }
 
 impl AssetBundle {
-    /// Auto-detect and load the bundle from a conventional location.
+    /// Load the bundle sitting next to the current executable.
     ///
-    /// Walks up from the current executable, checking each ancestor for an
-    /// `assets/manifest.toml`. This covers, without configuration:
+    /// The bundler writes to an `assets` directory beside the executable it
+    /// scanned, so `<exe_dir>/assets` is where a bundle belongs: `cargo run`
+    /// finds `target/<profile>/assets`, and a deployment ships the directory
+    /// alongside the binary.
     ///
-    /// - `<exe_dir>/assets/`: deployment, bundle shipped next to the binary.
-    /// - `target/<profile>/<bin>` -> `target/assets/` (typical `cargo run`).
-    /// - `target/<profile>/examples/<bin>` -> `target/assets/`.
-    /// - `target/<triple>/<profile>/<bin>` -> `target/assets/` (cross-compile).
-    /// - `target/<triple>/<profile>/examples/<bin>` -> `target/assets/`.
-    ///
-    /// The first directory that contains a readable `manifest.toml` is
-    /// loaded. The walk stops once it reaches a directory named `target`
-    /// (since `<target>/assets` is checked at that step) or after a bounded
-    /// number of ancestors. Returns [`io::ErrorKind::NotFound`] if no
-    /// candidate has a manifest.
-    ///
-    /// Use [`AssetBundle::load_dir`] when you already know the exact bundle
-    /// directory, such as a custom path passed to the asset bundler.
+    /// Use [`AssetBundle::load_dir`] when the bundle lives anywhere else, such
+    /// as a custom path passed to the asset bundler with `--out`.
     ///
     /// # Errors
     ///
-    /// Returns [`io::ErrorKind::NotFound`] if no candidate `assets` directory
-    /// contains a readable manifest, or propagates any I/O or parse error from
-    /// reading the located manifest via [`AssetBundle::load_dir`].
+    /// Returns [`io::ErrorKind::NotFound`] if there is no readable manifest
+    /// next to the executable, or propagates any I/O or parse error from
+    /// reading it via [`AssetBundle::load_dir`].
     pub fn load() -> io::Result<Self> {
         let exe = std::env::current_exe()?;
-        let exe_dir = exe
+        let dir = exe
             .parent()
             .ok_or_else(|| {
                 io::Error::new(
@@ -53,35 +42,16 @@ impl AssetBundle {
                     "current executable has no parent directory",
                 )
             })?
-            .to_path_buf();
+            .join("assets");
 
-        let mut tried = Vec::new();
-        let mut current = Some(exe_dir.as_path());
-        for _ in 0..6 {
-            let Some(d) = current else { break };
-            let candidate = d.join("assets");
-            if candidate.join(MANIFEST_NAME).is_file() {
-                return Self::load_dir(candidate);
-            }
-            tried.push(candidate);
-            if d.file_name() == Some(OsStr::new("target")) {
-                break;
-            }
-            current = d.parent();
+        if !dir.join(MANIFEST_NAME).is_file() {
+            return Err(io::Error::new(
+                io::ErrorKind::NotFound,
+                format!("no asset bundle at {}", dir.display()),
+            ));
         }
 
-        Err(io::Error::new(
-            io::ErrorKind::NotFound,
-            format!(
-                "no asset bundle found near {}: tried {}",
-                exe_dir.display(),
-                tried
-                    .iter()
-                    .map(|p| p.display().to_string())
-                    .collect::<Vec<_>>()
-                    .join(", ")
-            ),
-        ))
+        Self::load_dir(dir)
     }
 
     /// Load a bundle from a specific directory.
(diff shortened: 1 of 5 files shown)


JuanMarchetto:
Two builds of the same commit at the same profile produce different ids here, so the id is directory-scoped rather than profile-scoped. That changes what this fix covers.

Reproduced on Linux with the release profile on both sides: clone the same commit into two directories, run `topcoat asset bundle --release --out dist/assets` in each, and diff the manifests.

```
4c4
< id = 12679825631125907597
---
> id = 6092356416924241110
```

Serving one directory's binary with the other directory's bundle panics a worker thread on every request, and the client gets an empty reply:

```
thread 'tokio-rt-worker' panicked at topcoat-asset-0.5.0/src/config.rs:139:13:
failed to resolve asset Asset { id: AssetId(12679825631125907597) } in the asset catalog
```

`OUT_DIR` carries the absolute target path and the per-build hash, not only the profile name, so `asset!(concat!(env!("OUT_DIR"), "/tailwind.css"))` mints a fresh id in every checkout.

Bundling next to the scanned executable fixes the case this PR targets, and it fixes the two profiles clobbering each other's bundle. A bundle still cannot leave the machine that built it: build in CI, ship the binary and its `assets` directory to a server, and the ids no longer match. Deriving a generated asset id from the resolved file name and its content hash, rather than from the `OUT_DIR` path, would close that, and it is a separate change from this one.

crates/topcoat-asset/src/bundle.rs
```diff
@@ -23,11 +23,13 @@ impl AssetBundle {
...
-    /// - `target/<profile>/examples/<bin>` -> `target/assets/`.
-    /// - `target/<triple>/<profile>/<bin>` -> `target/assets/` (cross-compile).
-    /// - `target/<triple>/<profile>/examples/<bin>` -> `target/assets/`.
+    /// - `<exe_dir>/assets/`: where the bundler writes by default, next to the executable it
+    ///   scanned, and where a deployment ships the bundle alongside the binary. Since an asset's ID
+    ///   can depend on the cargo profile, each profile gets its own bundle here:
+    ///   `target/<profile>/assets/`, `target/<profile>/examples/assets/`, and their
+    ///   `target/<triple>/...` cross-compile counterparts.
+    /// - `target/assets/`, reached by walking up from any of those: bundles written to a custom
+    ///   `--out`, or by a Topcoat version that wrote one shared bundle for every profile.
```
pikaju: We don't want backwards compatibility

pikaju (changes requested):

crates/topcoat-asset/src/bundle.rs
AmeinEskinder: Dropped it. `AssetBundle::load` now only looks at `<exe_dir>/assets` instead of walking six ancestors, so the old shared `target/assets` is not found at all any more, and `asset clean` stopped probing for it too.

Net effect is a smaller diff than before: `load` loses about 40 lines, and a missing bundle names one directory rather than a list of candidates.

AmeinEskinder:
You're right, and my description was too narrow. `OUT_DIR` carries the absolute target path and the per-build hash, not just the profile name, so the id is scoped to the build directory rather than to the profile. I've corrected the wording in `asset.md` to say so.

It does mean this PR is narrower than it might look: it stops the two profiles from clobbering each other's bundle, and it turns a mismatch into a startup error naming the directory instead of a panic on the first request that renders the asset. It doesn't make a bundle portable between checkouts, so building in CI and shipping the binary plus its `assets` directory still won't line up.

Agreed that deriving a generated asset's id from the resolved file name and its content hash is the fix for that, and that it's a separate change. Worth its own issue?

pikaju (approved):

Issue #237, opened 2026-07-28
f64 text rendering differs between server and client


JuanMarchetto:
The browser f64 surrogate produces display text with `Number.prototype.toString` (`crates/topcoat-runtime/browser/src/surrogate/f64.ts` lines 59-73). The server renders the same value as `ViewPart::F64` through Rust's `Display` (`crates/topcoat-view/src/view.rs:259`, reached from `impl_primitive!(f64, push_f64, ref)` at `crates/topcoat-view/src/node.rs:72`). The two formats disagree on large magnitudes, small magnitudes, infinity and negative zero.

Repro:

```console
$ cat b2.rs
fn main() { for v in [1e21f64, 1e-7, f64::INFINITY, -0.0] { println!("{v}"); } }

$ rustc -o b2 b2.rs && ./b2
1000000000000000000000
0.0000001
inf
-0

$ node -e 'for (const v of [1e21, 1e-7, Infinity, -0.0]) console.log(v.toString())'
1e+21
1e-7
Infinity
0
```

An expression such as `$(x)` or `$(a / b)` renders one string in the server HTML and a different string after hydration or on the next shard update.

Fix direction: normalize the JavaScript formatting in the surrogate to match Rust `Display` output. #156 moved the server side to `Display` for parity with JavaScript, and these four cases remain. I can send the PR.

Will the planned reactive system overhaul replace the surrogate layer, so the fix lands in the right place?

Issue #125, opened 2026-07-18
Define and test Unicode semantics for shared String methods


srtdog64:
The Rust surrogate implements `len()` with Rust `str::len()`, which counts UTF-8 bytes, while the browser surrogate uses JavaScript `String.length`, which counts UTF-16 code units.
For example, `"한😊".len()` returns 7 on the server but 3 in the browser, so the same `expr!` expression can change meaning after hydration.
Could `len()` use the same Unicode rule on both sides and include a minimal non-ASCII parity test?

pikaju:
#126

PR #244, opened 2026-07-29
fix(runtime): match Rust semantics for string comparison and trim

### LINKED ISSUE #236: String comparison in reactive expressions disagrees between server and client
The browser string surrogate implements `lt`, `gt`, `le` and `ge` with JavaScript relational operators, which order by UTF-16 code unit (`crates/topcoat-runtime/browser/src/surrogate/string.ts` lines 18-32). The server evaluates the same operator through `StrSurrogate` (`crates/topcoat-runtime/src/surrogate/_str.rs` lines 30-46), which forwards to `str`'s `PartialOrd` and orders by UTF-8 bytes, which is code point order. `crates/topcoat-runtime/grammar/src/expr/expr_binary.rs:26` maps `BinOp::Lt` to the `lt` method on both sides, so the two evaluations of one `$()` expression can return opposite results whenever a character above U+FFFF is involved.

Repro:

```console
$ cat b1.rs
fn main() { println!("{}", "\u{FB00}" < "\u{1F600}"); }

$ rustc -o b1 b1.rs && ./b1
true

$ node -e 'const a = String.fromCodePoint(0xFB00), b = String.fromCodePoint(0x1F600); console.log(a < b, a > b);'
false true
```

U+FB00 sits below U+1F600 by code point, while its UTF-16 unit `0xFB00` sits above the high surrogate `0xD83D`, which reverses the browser result.

An expression such as `$(a < b)` renders one branch in the server HTML and the opposite branch after hydration or on the next shard update.

Fix direction: compare by code point in the surrogate, in the spirit of the `TextEncoder` change in #126. This is the same server/client divergence class as #125 / #126. I can send the PR.

Will the planned reactive system overhaul replace the surrogate layer, so the fix lands in the right place?

### LINKED ISSUE #238: trim() strips U+FEFF in the browser but not on the server
The browser string surrogate delegates `trim`, `trim_start` and `trim_end` to `String.prototype.trim`, `trimStart` and `trimEnd` (`crates/topcoat-runtime/browser/src/surrogate/string.ts` lines 46-56). ECMAScript counts U+FEFF as whitespace, so those methods strip it. The server runs `StrSurrogate::trim` and its siblings (`crates/topcoat-runtime/src/surrogate/_str.rs` lines 69-85), which forward to `str::trim`, whose whitespace set is the Unicode `White_Space` property and excludes U+FEFF. A string carrying a byte order mark keeps it on the server and loses it in the browser.

Repro:

```console
$ cat b3.rs
fn main() {
    let s = "\u{FEFF}x";
    println!("{:?} {:?}", s.trim(), s.trim_start());
}

$ rustc -o b3 b3.rs && ./b3
"\u{feff}x" "\u{feff}x"

$ node -e 'const s = String.fromCharCode(0xFEFF) + "x"; console.log(JSON.stringify(s.trim()), JSON.stringify(s.trimStart()));'
"x" "x"
```

An expression such as `$(s.trim())` renders the mark in the server HTML and drops it after hydration or on the next shard update, which also shifts any length, comparison or `is_empty` result derived from it.

Fix direction: align the surrogate with Rust's `White_Space` set instead of the ECMAScript whitespace set. I can send the PR.

Will the planned reactive system overhaul replace the surrogate layer, so the fix lands in the right place?

AmeinEskinder:
Closes #236. Closes #238.

Both are the same bug in the same file, and both are the part of #125 that #126 left undone: the browser string surrogate implements a shared-vocabulary method with different semantics than `str`, so one `$(...)` expression renders one thing in the server HTML and another after hydration or on the next shard update.

## Comparison (#236)

`gt`/`lt`/`ge`/`le` used JavaScript's relational operators, which order by UTF-16 code unit. `StrSurrogate` forwards to `str`'s `PartialOrd`, which orders by UTF-8 bytes — code point order. Any code point above U+FFFF compared against one in U+E000..U+FFFF disagrees, because the supplementary code point's high surrogate (0xD800-0xDBFF) sorts below it:

```
"\u{FB00}" < "\u{1F600}"     Rust: true     JavaScript: false
```

Now compared by code point.

## trim (#238)

`trim`/`trim_start`/`trim_end` delegated to `String.prototype`, which uses the ECMAScript whitespace set. Rust uses the Unicode `White_Space` property. The two differ **in both directions** — the issue names the first, but the second is just as live:

| code point | Rust `str::trim` | `String.prototype.trim` |
|---|---|---|
| U+FEFF ZWNBSP | keeps | **strips** |
| U+0085 NEL | **strips** | keeps |
| U+00A0 NBSP | strips | strips |
| U+180E | keeps | keeps |

Now trims the `White_Space` set. I checked the shipped `dist/index.js` regex against Rust on 14 probes across the set — U+0009, U+000B, U+000D, U+0020, U+0085, U+00A0, U+1680, U+2000, U+200A, U+2028, U+202F, U+205F, U+3000, and U+180E as a negative — and all agree.

## Tests

`string.test.ts` held only the #192 deref regression. #126's body says it added non-ASCII coverage, but its file list is `surrogate/string.ts` and `dist/index.js` only, so **the UTF-8 `len()` it fixed has never had a test**. This adds parity tests for all three: `len` in UTF-8 bytes, code point ordering (including the prefix/extension case), and the trim whitespace set on both ends. Expected values are what Rust produces for the same input.

`dist/index.js` is rebuilt, so the CI drift check passes.

## Docs

`expr.md` listed the shared vocabulary but documented none of its semantics, which is what #125 asked for and never got. Added one paragraph stating the rule these fixes follow: Rust defines the behavior and the JavaScript matches it, even where that departs from the JavaScript norm.

## Checks

`yarn build`, `yarn test` (12 passed), `git diff --exit-code -- dist/index.js`, `cargo test -p topcoat-runtime -p topcoat-runtime-macro`, `cargo clippy` on both crates.

## Note

#237 (f64 text rendering) is the same class but a different file and a different problem — matching Rust's `Display`, which never uses exponent notation, means expanding JS exponential form to full decimal (`1e308` is 309 digits, `5e-324` is 326). That deserves its own PR rather than riding along here.


JuanMarchetto:
I filed #236 and #238 and ran both repros against this branch, against rustc 1.97.1 for the Rust column.

Comparison, U+FB00 against U+1F600:

```
op                  rust     branch
lt(FB00, 1F600)     true     true
gt(FB00, 1F600)     false    false
le(FB00, 1F600)     true     true
ge(FB00, 1F600)     false    false
lt(1F600, FB00)     false    false
gt(1F600, FB00)     true     true
le(1F600, FB00)     false    false
ge(1F600, FB00)     true     true
```

Trim:

```
input             op            rust           branch
"\u{FEFF}x"       trim          "\u{FEFF}x"    "\u{FEFF}x"
"\u{FEFF}x"       trim_start    "\u{FEFF}x"    "\u{FEFF}x"
"x\u{FEFF}"       trim          "x\u{FEFF}"    "x\u{FEFF}"
"x\u{FEFF}"       trim_end      "x\u{FEFF}"    "x\u{FEFF}"
"  x  "           trim          "x"            "x"
"\u{85}x\u{85}"   trim          "x"            "x"
```

The same probes run against the minified comparison helper and the two trim regexes in `dist/index.js` return the same values, so the shipped bundle carries the fix. A randomized cross-check against rustc over 60000 comparison pairs, drawn from code points on both sides of the surrogate range, and 40000 trim inputs mixing the `White_Space` set with U+FEFF, U+180E, U+200B, U+2060 and payload characters, reported no mismatch. `yarn test` passes, 12 tests.

Every case the two issues report is covered.

pikaju (approved):
Thank you!

PR #245, opened 2026-07-29
fix(runtime): render f64 text the way Rust's Display does

### LINKED ISSUE #237: f64 text rendering differs between server and client

AmeinEskinder:
Closes #237.

The browser `F64` surrogate built its text with `Number.prototype.toString`, while the server renders `ViewPart::F64` through Rust's `Display`. Same expression, two different strings, so the page changes after hydration.

## What differs

The issue reports the first three rows. The fourth came out of comparing the two implementations over a large sample.

| | Rust `Display` | `Number.prototype.toString` |
|---|---|---|
| `1e21` | `1000000000000000000000` | `1e+21` |
| `1e-7` | `0.0000001` | `1e-7` |
| `f64::INFINITY` | `inf` | `Infinity` |
| `-0.0` | `-0` | `0` |
| `1690060720831323.25` | `1690060720831323.3` | `1690060720831323.2` |

That last one is a tie-break disagreement. Both sides emit the shortest digits that round-trip, but when two equally short strings are *exactly* as close to the value, ECMAScript picks the one ending in an even digit and Rust picks the one further from zero. `NaN` already agreed.

## The fix

Rust never uses exponential notation, so the bulk of it is expanding JavaScript's exponential form into positional decimal — exact at any magnitude, including the 309-digit `1e308` and the 326-character `5e-324`.

The tie is detected exactly rather than heuristically: `v` is decomposed into the `mantissa * 2 ** exponent` it is actually stored as, and the decimal midpoint `(2 * digits + 1) * 10 ** scale` is compared against it with BigInt, both sides scaled until every exponent is non-negative. An odd final digit rules the tie out before any of that runs, since ECMAScript only lands on an even one by resolving a tie.

## Verification

I generated Rust `Display` output for **300,016 values** (236,016 distinct doubles) and compared every one:

- the fixed cases above, plus `f64::MIN_POSITIVE`, `f64::EPSILON`, `f64::MAX`, subnormals
- 60,000 random bit patterns and uniform values
- a quarter/eighth-step sweep at magnitudes where ties cluster

**All 300,016 agree.** 8% of that sample exercises the tie correction, so the branch is well covered rather than incidental. The committed `f64.test.ts` keeps a readable subset of those cases; the bulk-comparison harness was a scratch tool and is not in the diff.

## Checks

`yarn build` (tsup + tsc), `yarn test` (14 passed), `cargo test -p topcoat-runtime-macro --doc`, `cargo clippy -p topcoat-runtime`. `dist/index.js` is rebuilt so the drift check passes.

## Note

`expr.md` gains one clause on the `f64` bullet. #244 (#236 + #238) adds a paragraph further down the same section; the two hunks are far enough apart not to conflict, but they are worth reading together since they state the same rule — the shared vocabulary is defined by Rust and the JavaScript matches it.


JuanMarchetto:
I filed #237 and ran the repro against this branch, against rustc 1.97.1 for the Rust column.

```
value                 rust Display              branch toNodeText
1e21                  1000000000000000000000    1000000000000000000000
1e-7                  0.0000001                 0.0000001
f64::INFINITY         inf                       inf
-0.0                  -0                        -0
f64::NEG_INFINITY     -inf                      -inf
0.0                   0                         0
f64::NAN              NaN                       NaN
1690060720831323.25   1690060720831323.3        1690060720831323.3
1e308                 309 chars, no exponent    identical, 309 chars
5e-324                326 chars, no exponent    identical, 326 chars
```

`toAttributeValue` returns the same string as `toNodeText` on every row. The same values run against the minified `display` in `dist/index.js` return the same strings, so the shipped bundle carries the fix. A cross-check against rustc over 204023 values, covering 100000 random bit patterns, 100000 random values across magnitudes 1e-20 to 1e20, quarter-step and eighth-step sweeps where ties cluster, and `MIN_POSITIVE`, `EPSILON`, `MAX`, `MIN` and subnormals, reported no mismatch.

Every case the issue reports is covered.

pikaju:
Thank you, could you resolve the merge conflicts?

AmeinEskinder:
Rebased onto main. The only conflict was `dist/index.js`, since #244 rebuilt the same minified bundle; I resolved it by re-running `yarn build` over the merged sources rather than taking either side, so the bundle carries both changes. `expr.md` merged cleanly, the two hunks are far enough apart.

`yarn test` passes, 19 tests, and `git diff --exit-code -- dist/index.js` after a fresh build is clean.

pikaju (approved):

pikaju:
Thank you!

PR #247, opened 2026-07-29
fix(runtime): reject non-2xx shard responses instead of rendering them

JuanMarchetto:
## Summary

`ReactiveScope.fetchAndReplace` passes `res.text()` into the shard's content without reading the response status, so any non-2xx response becomes the shard's new content. A 404 or a 405 carries an empty body, so the rendered content is deleted while the rest of the page keeps working, and the only trace is the browser network log.

I hit this in an application I built on Topcoat. A restart moved the shard endpoint, the open tab's next POST to the old URL answered `405` with `content-length: 0`, and fifteen rows of request history were removed from the page while the heading and the filter box above them stayed. The swap is independent of how the shard id is minted: a shard whose endpoint returns 500 in production erases itself the same way.

The patch throws on a non-ok response, before the swap. `Procedure.call` in `surrogate/procedure.ts` does the same for the sibling endpoint, and the existing `catch` in `fetchAndReplace` rethrows anything that is not an abort, so the failure surfaces the way a network failure does and the rendered content stays on the page.

The test drives `fetchAndReplace` against a stubbed 405 and asserts that no child is removed. Against unpatched `scope.ts` it fails on that assertion, with the content node in the removed list:

```
FAIL  src/scope.test.ts > keeps the rendered content when the shard responds with an error
AssertionError: expected [ { nextSibling: {} } ] to deeply equal []
```

This regenerates `crates/topcoat-runtime/browser/dist/index.js`, which #244 and #245 also change.

## Testing

Run at `be29741`.

```
$ cargo +nightly fmt --all
(no output, exit 0)

$ cargo topcoat fmt
formatted 521 files (0 modified), 5 failed in 3s
```

The 5 failures are the Leptos benchmark sources.

```
$ cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 06s

$ cargo test --workspace --all-features
115 `test result: ok` lines, 1643 passed, 0 failed, exit 0

$ RUSTDOCFLAGS="--cfg docsrs -Dwarnings" cargo +nightly doc --workspace --all-features --no-deps --locked
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 20.31s
   Generated target/doc/alpine_ajax/index.html and 59 other files
```

```
$ cd crates/topcoat-runtime/browser
$ yarn install --frozen-lockfile
Done in 18.91s.

$ yarn build
ESM dist/index.js     17.72 KB
ESM Build success in 37ms

$ git diff --exit-code -- dist/index.js
(clean against the committed bundle, exit 0)

$ yarn test
 Test Files  4 passed (4)
      Tests  8 passed (8)
```

## Disclaimer


pikaju:
Hey, could you fix the merge conflict? Thank you!

JuanMarchetto:
Rebased on main with #245 in it. The conflict was the generated bundle, so I rebuilt it rather than merging the minified output; `git diff --exit-code` on `dist/index.js` is clean against a fresh `yarn build`. Checks rerun on the new head: fmt, clippy with all features, the workspace tests, and `yarn test` at 20 passing.

pikaju (approved):

PR #249, opened 2026-07-29
fix(asset): register one route per bundled file

JuanMarchetto:
## Summary

`asset!` mints an `AssetId` per call site and the bundled filename derives from the file's content hash, so declaring one file from two modules yields two catalog entries pointing at the same bundled file. `RouterBuilderAssetExt::assets` registered one `AssetRoute` per catalog entry, so `RouterBuilder::build` saw two routes for one URL and panicked before the application bound its port:

```
thread 'main' panicked at crates/topcoat-router/src/router.rs:541:25:
duplicate route registered for `GET /_topcoat/assets/mark-16313d41ffdefca4.svg`
```

The panic names the URL and no source location. I hit it in an application I built on Topcoat, where a logo was declared in two modules.

The fix has two halves.

**Router.** The `Host::Serve` arm registers a route the first time it sees a bundled filename. The `Content-Type` is not part of the key. A bundled filename carries a content hash, so entries that share a filename share their bytes: same file, same URL, same content, and collapsing them does not change what is served.

**Bundler.** `Bundler::bundle` now records the `Content-Type` of every output filename it prepares and fails with `BundleError::ConflictingContentTypes` when a later asset resolves to a filename already prepared under a different type. The message names the output filename, both types, and the `rename` that separates them:

```
conflicting content types "image/svg+xml" and "text/plain" for bundled file mark-16313d41ffdefca4.svg: serving one file as two content types needs a different `rename` on one of the declarations
```

That is the contradiction the filename-only key would otherwise resolve in silence: two entries claim different types for one route, `AssetCatalog::assets` iterates a `HashMap`, and the surviving type is arbitrary. The bundle now fails before a manifest is written, so the catalog the router loads can no longer contain it.

Of the four `AssetOptions` fields, `content_type` is the only one that can differ between two entries that share an output filename and change how the file is served. `rename` and `extension` shape the filename itself, so entries that share a filename agree on their effect; `checksum` is asserted against the source bytes during bundling and reaches neither the manifest nor the route.

The alternative fix is to intern `AssetId`s by resolved path in the bundler, so one file has one id. I did not take it because it changes asset identity semantics, while this one is local to route registration and manifest assembly.

Two open PRs sit next to this: #196 edits the same `Host::Serve` loop to collect static file URLs, and #243 changes where the bundler writes the bundle, while the new check sits in the `Bundler::bundle` loop that collects prepared entries.

## Testing

`one_file_declared_at_two_call_sites_serves_one_route` covers the defect: it panics with `duplicate route registered for GET /_topcoat/assets/mark-16313d41ffdefca4.svg` at 01cdbd9 and passes with the fix, serving `200` and `image/svg+xml` from the single surviving route.

`a_catalog_that_disagrees_on_content_type_still_serves_one_route` pins the filename-only key: two entries share the bundled filename and declare different content types, and it panics with the same `duplicate route registered` message under the previous key at 919a533. This is the test that was `one_file_declared_with_two_content_types_serves_one_route`. The bundler no longer produces such a catalog, but `Manifest::parse` accepts one built elsewhere or embedded ahead of time, as on WebAssembly, so the router still has to tolerate it. The test asserts the served type is one of the two declared and does not pin which.

`one_output_file_declared_with_two_content_types_is_rejected` covers the new check: two declarations of one file with different `content_type` options, expecting `BundleError::ConflictingContentTypes` carrying the output filename and both types. `bundle` returns `Ok` at 40b8c23, so the test fails there and passes now.

`one_output_file_declared_with_one_content_type_is_accepted` pins the other side: two declarations that resolve to one output filename and agree on the type still bundle, into two manifest entries pointing at one file.

- `cargo +nightly fmt --all`, then `--check`: exit 0
- `cargo topcoat fmt`: 541 files, 0 modified (the 5 failures are the Leptos benchmark files)
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`: exit 0
- `cargo test --workspace --all-features`: 115 `test result: ok` lines, 0 failed; `topcoat-asset` reports `32 passed; 0 failed`
- `RUSTDOCFLAGS="--cfg docsrs -Dwarnings" cargo +nightly doc --workspace --all-features --no-deps --locked`: exit 0

## Disclaimer

### DIFF
--- a/crates/topcoat-asset/src/router.rs
+++ b/crates/topcoat-asset/src/router.rs
@@ -80,7 +80,7 @@ impl RouterBuilderAssetExt for RouterBuilder {
                 let mut builder = self;
                 let mut registered = HashSet::new();
                 for asset in config.catalog.assets() {
-                    if registered.insert((asset.name(), asset.content_type())) {
+                    if registered.insert(asset.name()) {
                         builder = builder.route(AssetRoute::new(dir, asset));
                     }
                 }
@@ -157,11 +157,16 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "duplicate route registered")]
-    fn one_file_declared_with_two_content_types_conflicts() {
+    fn one_file_declared_with_two_content_types_serves_one_route() {
         let dir = temp_dir("conflicting-content-types");
+        fs::write(dir.join(FILE), "<svg></svg>").unwrap();
         let bundle = bundle(dir, &[(MAIN, "image/svg+xml"), (OTHER, "text/plain")]);
 
-        let _ = Router::builder().assets(bundle).build();
+        let router = Router::builder().assets(bundle).build();
+
+        let parts = get(&router, &format!("{ASSET_ROUTE_PREFIX}/{FILE}"));
+        assert_eq!(parts.status, StatusCode::OK);
+        let content_type = parts.headers.get(CONTENT_TYPE).unwrap();
+        assert!(content_type == "image/svg+xml" || content_type == "text/plain");
     }
 }
--- a/crates/topcoat-asset/src/router.rs
+++ b/crates/topcoat-asset/src/router.rs
@@ -157,7 +157,7 @@ mod tests {
     }
 
     #[test]
-    fn one_file_declared_with_two_content_types_serves_one_route() {
+    fn a_catalog_that_disagrees_on_content_type_still_serves_one_route() {
         let dir = temp_dir("conflicting-content-types");
         fs::write(dir.join(FILE), "<svg></svg>").unwrap();
         let bundle = bundle(dir, &[(MAIN, "image/svg+xml"), (OTHER, "text/plain")]);
--- a/crates/topcoat-asset/src/router.rs
+++ b/crates/topcoat-asset/src/router.rs
@@ -2,6 +2,9 @@
 // cannot resolve; they degrade to plain text instead of failing the build.
 #![cfg_attr(not(feature = "serve"), allow(rustdoc::broken_intra_doc_links))]
 
+#[cfg(feature = "serve")]
+use std::collections::HashSet;
+
 use topcoat_router::RouterBuilder;
 
 #[cfg(feature = "serve")]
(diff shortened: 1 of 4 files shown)


crates/topcoat-asset/src/router.rs
```diff
@@ -74,8 +78,11 @@ impl RouterBuilderAssetExt for RouterBuilder {
             #[cfg(feature = "serve")]
             Host::Serve { dir } => {
                 let mut builder = self;
+                let mut registered = HashSet::new();
                 for asset in config.catalog.assets() {
-                    builder = builder.route(AssetRoute::new(dir, asset));
+                    if registered.insert((asset.name(), asset.content_type())) {
```
pikaju: I think the `content_type` should not be used as a key in the set, because two assets with the same name would still collide in the routes

pikaju (changes requested):

JuanMarchetto:
The set now keys on the bundled filename, not on the pair with `Content-Type`, so two entries sharing a filename register one route; `one_file_declared_with_two_content_types_serves_one_route` covers that case and fails under the old key. One judgement call: when the two entries declare different content types, the surviving route takes the type of whichever entry the catalog yields, and the catalog iterates in arbitrary order. I can make that contradiction panic with a message naming the file and both types if you want it reported instead of resolved.

pikaju:
@JuanMarchetto we need one more check: the bundler should reject if two assets have the same output file name (i.e. same name + hash + extension) BUT different configurations (e.g. different content type) as this would create a silent issue right now. If the developer wants the same file with different content types they should be forced to specify a different asset name as well.

JuanMarchetto:
The bundler rejects two assets that resolve to one output filename with different content types, and the error names the file, both types, and the `rename` that separates them. The check sits in the `Bundler::bundle` loop that collects prepared entries, where it sees every asset destined for one output file before the manifest is written, and `one_output_file_declared_with_two_content_types_is_rejected` pins it. The router keeps the filename-only route key, and its content-type test became `a_catalog_that_disagrees_on_content_type_still_serves_one_route`, since a manifest built elsewhere can still hand the router that catalog.

One sequencing note: #243 changes where the bundler writes the bundle, and this check lands inside `Bundler::bundle`.

pikaju (approved):

pikaju:
Thank you!

PR #253, opened 2026-07-29
fix(runtime): derive shard and procedure ids from the declaration site

### LINKED ISSUE #252: Shard and procedure endpoint ids change on every build
`#[shard]` and `#[procedure]` mint their endpoint id with `Uuid::new_v4()` at macro expansion time, so every compilation produces new URLs. A browser tab left open across a deploy keeps polling the previous binary's id, and the new binary does not serve it.

Four builds of one unchanged source in an application I built produced four ids:

```
d646e935-433f-44a3-88ea-fe71155e2a28
68c1a3c6-...
a91567b3-...
65adb204-...
```

What the open tab does then, from the network log of a tab that was never reloaded:

```
POST /_topcoat/shards/a823c355-9fd9-438f-a470-0507653715a8 -> 200 (3ms, 969B)
POST /_topcoat/shards/a823c355-9fd9-438f-a470-0507653715a8 -> 200 (2ms, 963B)
... rebuild and restart ...
POST /_topcoat/shards/a823c355-9fd9-438f-a470-0507653715a8 -> 405 (0B)
```

The status is 405 rather than 404 when a catch-all page claims the path. The failure is silent in both cases: `fetchAndReplace` swaps the empty body in, so the fifteen rows on screen are deleted while the heading, the summary and the filter box outside the shard stay put. Nothing reaches the console and no request fails from the application's point of view. A control run confirms the rebuild causes it and not the restart: stopping and starting the same binary cost the open tab one refused connection, after which it kept polling the same URL and picked up a delivery four seconds later.

@AmeinEskinder reproduced this independently on `examples/shard`, three builds with only a comment changed, and found two things I had not:

- `#[procedure]` has the same construction at `procedure.rs:123`, so a stale tab's button click fails the same way.
- `topcoat asset bundle` compiles the crate, so the bundling step can mint fresh ids partway through a deploy sequence.

There is already a pattern for a stable opaque id in the tree: `AssetId::new` hashes the crate name, the source file and the asset path. Something equivalent over crate name, module path and ident would survive a rebuild while staying opaque.

One argument against determinism, which belongs in the decision rather than in a reviewer's head: a fresh id per build means a URL someone learned once stops working at the next deploy, and making ids deterministic makes that URL permanent. It is not protection either way, since the page hands the id to anyone who loads it, and shard endpoints need their own authorization regardless.

Related: #247 covers the swap that makes the failure silent, and the id churn is the other half of what an open tab experiences across a deploy.

JuanMarchetto:
## Summary

Fixes #252. `#[shard]` and `#[procedure]` minted their endpoint id with `Uuid::new_v4()` at macro expansion time, so every compilation produced new URLs under `/_topcoat/shards/<id>` and `/_topcoat/procedures/<id>`. A browser tab left open across a deploy keeps polling the previous binary's id, the new binary does not serve it, and the failure is silent: the runtime swaps the empty body in and the widget content disappears with nothing in the console. `topcoat asset bundle` compiles the crate, so the bundling step can mint fresh ids partway through a deploy sequence.

The macros now emit `module_path!()` into the generated code and hash it in the declaring crate, the same shape as `AssetId::new` doing its work at the declaration site. The key is `CARGO_CRATE_NAME`, the module path, and the function name, folded with the in-tree `fnv1a` const hash and rendered as 16 ASCII hex characters. Two items cannot share a name inside one module, so that key is collision-free for declarations at module scope, where hashing the source file would not be: two modules in one file can each hold a `#[shard] fn rows`. The derivation, the argument for it, and the check that it compiles and survives a rebuild are AmeinEskinder's, from his comment on #252.

`ShardId` and `ProcedureId` keep their `&'static str` newtype and `const fn new`, and the route path still comes from `as_str`. Nothing in the tree parses or validates the id shape: the two route builders interpolate it with `format!`, `ReactiveScope` writes it into the scope comment, and the browser runtime passes it through `encodeURIComponent`. The id is 16 hex characters rather than a 36-character uuid, which no consumer notices.

The tradeoff, stated plainly because it is a maintainer decision: a derived id makes a shard or procedure URL permanent across deploys, where today it changes on every build, so someone who records a URL keeps a working one instead of losing it at the next deploy. Shard endpoints need their own authorization either way, as the guards note from #251 says. If you want a different derivation, such as a per-deploy salt or an opt-in `#[shard(id = "...")]` escape hatch, redirect me and I will redo it.

## Testing

Before, on `main`, expanding `examples/shard` twice with only a comment changed between the two runs:

```
$ cargo +nightly rustc -p shard --profile=check -- -Zunpretty=expanded | grep -o 'ShardId::new("[^"]*")'
ShardId::new("4d223daa-a583-4c0a-9976-fb7ce4c1b9c7")

$ printf '\n// touch to force a rebuild\n' >> examples/shard/src/main.rs
$ cargo +nightly rustc -p shard --profile=check -- -Zunpretty=expanded | grep -o 'ShardId::new("[^"]*")'
ShardId::new("5404c1da-f715-45ef-927c-fa05f6a46b92")
```

After, the id is computed in the declaring crate, so it appears in the binary rather than in the expansion. `6aaf9450fdc57264` is `endpoint_id_hash("shard", "shard", "combobox_content")`, and the same comment edit leaves it in place:

```
$ cargo build -p shard && grep -ao 6aaf9450fdc57264 target/debug/shard
6aaf9450fdc57264
$ md5sum target/debug/shard
eadc9c7ad4d30bc55034f7fd3da6427e  target/debug/shard

$ printf '\n// touch to force a rebuild\n' >> examples/shard/src/main.rs
$ cargo build -p shard && grep -ao 6aaf9450fdc57264 target/debug/shard
6aaf9450fdc57264
$ md5sum target/debug/shard
eadc9c7ad4d30bc55034f7fd3da6427e  target/debug/shard
```

`examples/procedure` behaves the same way, serving `82edc9c5c31a7706` for `print_on_server`.

New tests: `crates/topcoat-runtime/src/id.rs` pins the id of a known crate, module, and name, so a change in the derivation fails loudly, and covers the module and crate separation plus the hex width. The shard and procedure grammars each gain a test that two expansions of one source are identical and that the id comes from crate, module, and name.

Check gauntlet from `.agents/skills/check/SKILL.md`:

```
$ cargo +nightly fmt --all
$ cargo +nightly fmt --all --check
(clean)

$ cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 33s
(exit 0, no warnings)

$ cargo test --workspace --all-features
(exit 0, 1656 passed, 0 failed)

$ RUSTDOCFLAGS="--cfg docsrs -Dwarnings" cargo +nightly doc --workspace --all-features --no-deps --locked
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 24.29s
   Generated target/doc/alpine_ajax/index.html and 59 other files
(exit 0)
```

`crates/topcoat-runtime/browser` is untouched, so `dist/index.js` is unchanged and the yarn steps do not apply.

## Disclaimer

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3128,7 +3128,6 @@ dependencies = [
  "syn",
  "topcoat-core-grammar",
  "topcoat-runtime",
- "uuid",
 ]
 
 [[package]]
--- a/crates/topcoat-runtime/grammar/Cargo.toml
+++ b/crates/topcoat-runtime/grammar/Cargo.toml
@@ -23,7 +23,6 @@ syn = { workspace = true, features = ["full", "visit"] }
 
 serde.workspace = true
 serde_json.workspace = true
-uuid = { workspace = true, features = ["v4"] }
 
 [lints]
 workspace = true
--- a/crates/topcoat-runtime/grammar/src/procedure.rs
+++ b/crates/topcoat-runtime/grammar/src/procedure.rs
@@ -120,13 +120,29 @@ impl ToTokens for Procedure {
         };
         let return_ty = quote! { <#return_ty as #topcoat_internal::ResultExt>::T };
 
-        let id = uuid::Uuid::new_v4().to_string();
+        let id_ident = format_ident!("__TOPCOAT_PROCEDURE_ID_{}", ident);
 
         quote! {
+            #[doc(hidden)]
+            #[allow(non_upper_case_globals)]
+            const #id_ident: &'static ::core::primitive::str = {
+                const HASH: u64 = #topcoat_runtime::endpoint_id_hash(
+                    ::core::env!("CARGO_CRATE_NAME"),
+                    ::core::module_path!(),
+                    ::core::stringify!(#ident),
+                );
+                const BYTES: [u8; #topcoat_runtime::ENDPOINT_ID_LEN] =
+                    #topcoat_runtime::endpoint_id_hex(HASH);
+                match ::core::str::from_utf8(&BYTES) {
+                    ::core::result::Result::Ok(id) => id,
+                    ::core::result::Result::Err(_) => ::core::panic!("hex is ascii"),
+                }
+            };
+
             #(#docs)*
             #[allow(non_upper_case_globals)]
             #vis const #ident: &#topcoat_runtime::Procedure::<(#(#arg_tys,)*), #return_ty> = &#topcoat_runtime::Procedure::new(
-                #topcoat_runtime::ProcedureId::new(#id),
+                #topcoat_runtime::ProcedureId::new(#id_ident),
                 |cx, body| {
                     #[allow(clippy::unused_async)]
                     #item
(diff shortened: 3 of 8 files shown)


pikaju:
Thank you, we have a solution for this now in the latest release. You can specify the path for your shard/procedure manually. I think this is more robust than deriving it from the code location.

PR #259, opened 2026-07-29
fix(router): reject mismatched layer parameter names

carllerche:
Reject mismatched parameter names at `.build()`:

```text
layer: /accounts/{account_id}
route: /accounts/{id}
```

Most of the diff is threading the definition location of the route and layer for better error messages.

Tested with workspace checks.

AI: Codex (GPT-5) implemented and tested this change.


pikaju:
I added a simpler, more general check at #300

PR #260, opened 2026-07-30
fix(runtime): unwrap surrogates before writing DOM properties

SixSeven-Labs:
ran into this building a little catalog browser on 0.5.0. two checkboxes bound with `:checked` to `false` signals, both came up checked.

`binding.ts` writes value/checked/selected/indeterminate as properties instead of attributes, which is right, but it hands the computed value straight over:

```ts
(el as Element & Record<string, unknown>)[name] = value;
```

that value is a surrogate. objects are truthy, so `el.checked = Bool(false)` checks the box. and the attribute path right below it does the correct thing and calls removeAttribute, so devtools shows no `checked` attribute while the property disagrees.  took me a minute to find.

`value` gets away with it by accident. assigning an object goes through toString, and `Str.toString` hands back the string we wanted anyway.

fix moves the unwrap into `propertyValue(name, value)`. boolean props read `isAttributePresent()`, `value` reads `toAttributeValue()`, or `""` when the value is absent since `None.toAttributeValue()` throws. non surrogate values pass through, coerced to bool for the boolean props so `raw!` keeps working.

checked against `examples/runtime` with a few extra inputs on the show page, where `off` is `:checked=$(show.get())` and show starts false:

```
before: {off: true,  on: true, text: "hidden"}
after:  {off: false, on: true, text: "hidden"}
```

test is a unit test on `propertyValue`, no dom needed. dist rebuilt with `npm run build`.


pikaju:
Sorry for the wait, I believe this is fixed now. Thank you for the PR though!

PR #261, opened 2026-07-30
fix(router): restrict WebSocket origins

carllerche:
## Summary

- Reject browser WebSocket handshakes whose `Origin` does not match the configured policy. The upgrade path previously accepted every browser origin.
- Default the router policy to the `base_url` origin. An explicit `websocket_origins(...)` replaces that default, endpoint `allow_origin(...)` extends it, and public sockets can opt out explicitly.
- Document the policy, update the WebSocket example, and add unit plus real-connection coverage.

## Testing

- `cargo +nightly fmt --all`
- `cargo topcoat fmt` (expected Leptos benchmark parse errors only)

## AI disclaimer

This PR was created with Codex (GPT-5), which implemented the change, updated tests and documentation, and ran the local verification commands.


pikaju:
implemented more generally by #276

PR #267, opened 2026-07-30
feat(router): add a too-many-requests error with a Retry-After hint

AmeinEskinder:
Sibling to #266, kept separate so each stays small.

## Summary

`error_into_response` maps the router's own error types onto status codes through a closed list of downcasts and turns everything else into a 500. The module has no constructor for 429, so a rate limiter has nothing to return and a refused request answers as though the server were broken.

429 and 503 are deliberately different answers. 429 refuses a *caller* for how much they asked; 503 says the *server* is at capacity for everyone. A client that can tell them apart throttles its own traffic in the first case and fails over in the second — answering 500 for either tells it to do neither, and tells a load balancer to pull a healthy instance.

## The change

`too_many_requests(secs)` responds 429 and sets `Retry-After`, so the refusal says both what the caller did and when to come back. Without the header a rate limiter teaches callers nothing except to retry immediately, which is the behaviour a limiter exists to prevent.

Same shape as the existing constructors, added to the downcast list and the docs' constructor list.

## Testing

Three tests cover the type: the status and header, the accessor, and the `Display` string.

The one that matters covers the mapping. Removing the new arm from the downcast list leaves the constructor and its own `IntoResponse` passing while the handler reports:

```
assertion `left == right` failed
  left: 500
 right: 429
```

`an_error_that_is_not_on_the_list_still_maps_to_500` pins the fallthrough so the guard cannot be satisfied by making everything 429.

## Checks

`cargo fmt --all --check`, `cargo clippy -p topcoat-router --all-targets --all-features` (0 warnings), `cargo test -p topcoat-router` (303 + 71 passed), doctest.

## Note on merge order

This and #266 both add a `mod`, a `pub use`, a downcast arm, and a docs sentence in the same places, so whichever lands second needs a small rebase — the conflict is textual adjacency, not disagreement. Happy to rebase this one onto #266, or to fold them together if you would rather review one PR than two.

### DIFF
--- a/crates/topcoat-router/docs/error.md
+++ b/crates/topcoat-router/docs/error.md
@@ -4,7 +4,7 @@ Every page, layout, layer, and route handler returns a `Result`. An `Err` become
 
 # Constructors
 
-Every error type in this module has a constructor function named after its response. For example, [`not_found()`](not_found) responds 404 with [`NotFoundError`], [`redirect(uri)`](redirect) responds 307 with [`RedirectError`], [`bad_request(description)`](bad_request) responds 400 with [`BadRequestError`] and a client-safe description, and [`service_unavailable(secs)`](service_unavailable) responds 503 with [`ServiceUnavailableError`] and a `Retry-After` header.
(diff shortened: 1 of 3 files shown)


pikaju:
Could you fix the conflicts?

pikaju (dismissed):

pikaju:
@AmeinEskinder more conflicts :(

AmeinEskinder:
Chief @pikaju addressed.

pikaju (approved):
Thank you!

PR #268, opened 2026-07-30
feat(router): add Js and Wasm response wrappers

AmeinEskinder:
## Summary

`Css` already wraps a body and sets its media type for a route that serves a stylesheet by hand. Scripts and WebAssembly modules had no equivalent, so serving either meant attaching the header through a tuple at every call site.

Those two are worth a wrapper each because their media type is **checked, not sniffed**:

- A browser refuses to execute a `<script type="module">` whose response does not carry a JavaScript media type. It surfaces as a MIME type mismatch in the console rather than a script error, which is a confusing way to find out.
- `WebAssembly.compileStreaming` and `instantiateStreaming` reject any response that is not exactly `application/wasm`. A module served as `application/octet-stream` fails to instantiate with perfectly correct bytes.

Both come out of serving a hand-rolled script and a wasm engine from routes rather than through the asset bundle.

## Shape

`Js<T>` sets `text/javascript; charset=utf-8`, `Wasm<T>` sets `application/wasm`, both over any `T: Into<Body>` — the same shape as `Css`'s response half.

They are response-only. Neither is a body a server reads from a request, so there is no extractor half to mirror, and adding one would be surface with no caller.

## Testing

One test each for the status quo: the media type set, and the body passed through unchanged. `Wasm` gets a second one asserting the header carries no parameters, since a `; charset=...` would break the streaming entry points exactly the way the wrong type does — that is the failure the wrapper exists to prevent, so it is worth pinning rather than assuming.

## Checks

`cargo fmt --all --check`, `cargo clippy -p topcoat-router --all-targets --all-features` (0 warnings), `cargo test -p topcoat-router` (301 + 72 passed), doctests for both wrappers.

### DIFF
--- a/crates/topcoat-router/docs/content.md
+++ b/crates/topcoat-router/docs/content.md
@@ -62,6 +62,8 @@ async fn create_user() -> Result<(StatusCode, Json<User>)> {
 }
 ```
 
+[`Js`] and [`Wasm`] are response-only wrappers for the two media types a browser checks rather than guesses: it refuses to execute a `<script type="module">` that does not arrive as JavaScript, and `WebAssembly.compileStreaming` rejects anything that is not exactly `application/wasm`. Reach for them when a route serves a script or a module by hand rather than through the asset bundle.
+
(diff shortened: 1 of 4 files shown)


pikaju (dismissed):

pikaju:
Could you fix the CI issues?

pikaju (approved):

PR #274, opened 2026-07-31
fix(view): allow keyword element names

ShiroKSH:
## Summary

- accept Rust keywords when parsing static element names
- cover the SVG `<use>` element through the full `view!` render path

The parser already uses `Ident::parse_any` after lookahead, but the lookahead only accepted non-keyword identifiers. That made valid SVG markup fail before parsing.

## Testing

- `cargo +nightly fmt --all`
- formatter CLI repro with `<svg><use href="#icon"></use></svg>`


pikaju (approved):
Thank you!

PR #277, opened 2026-07-31
Fix comment formatting in tailwind.md

Lucas-BRT:


pikaju:
Why is this needed?

Lucas-BRT:
Sorry, I created the PR directly from the GitHub page and forgot to add the PR description.


The original code snippet in the docs says to add this to build.rs:

```rust
# #[allow(clippy::needless_doctest_main)]
fn main() {
    topcoat::tailwind::BuildConfig::new().render().unwrap();
}
```

But the extra # before the `#[allow(clippy::needless_doctest_main)]` causes the following error:

```bash
lucasbrt@pop-os:~/Projects/top$ cargo build
   Compiling top v0.1.0 (/home/lucasbrt/Projects/top)
error: expected one of `!` or `[`, found `#`
 --> build.rs:1:3
  |
1 | # #[allow(clippy::needless_doctest_main)]
  |   ^ expected one of `!` or `[`

error: could not compile `top` (build script) due to 1 previous error
```

Removing the # solves the issue.

Probably is a typo during the writing of the documentation, so this PR is more for better consistency of the documentation.

pikaju:
Ah, @Lucas-BRT the extra `#` is actually a marker for Rust and [docs.rs](https://docs.rs/topcoat) to hide the line of code. You've probably been reading it either on GitHub or in the source code, but if you look up the documentation on docs.rs it wont show up.

I think it's better to leave it as is, the user isn't supposed to copy this line into their code (and you do not need to either), it just disables a documentation warning.

PR #292, opened 2026-08-02
feat(view): render component trees concurrently

petehunt:
## Summary

`view!` now collects static segments and unresolved component futures into an internal tree, resolves the pending nodes concurrently, and composes their output in source order.

This extends concurrency across nested markup, selected `if`/`match` branches, loop iterations, and component boundaries. Child content is passed to its parent through an internal placeholder, so independent components in parent, child, and deeper descendant views can all start before any completes. Unselected branches do not run, and views without components keep the existing direct `ViewParts` path.

The implementation uses structured futures rather than spawned tasks: dropping the render future drops its pending component futures. The placeholder is filled before the completed tree is returned for rendering.

## Comparison with the focused alternative

This is an independent, more general alternative to [#284](https://github.com/tokio-rs/topcoat/pull/284), which only joins peer components within safe emitter groups.

| | #284 | This PR |
|---|---|---|
| Concurrency | Peer components within an emitter group | Components throughout the selected tree, including loop iterations and nested component boundaries |
| Boundaries | Rust expressions, bindings, statements, and control flow end a group | Selected control flow becomes part of the concurrent tree |
| Runtime changes | None to `View`; concurrency is emitted as local joins | Adds runtime tree nodes, boxed futures, and child-view placeholders |
| Component children | Remain completed `View` values before the parent starts | Use placeholders so parent and child work can overlap |
| Compatibility | Smaller timing change; no new `Send` requirement | Broader side-effect timing changes; pending tree futures must be `Send`; synchronously rendering a child inside its component is incompatible with an unresolved placeholder |
| Cost | Generated tuple joins for each group | Allocation for pending nodes plus placeholder synchronization |

#284 favors a focused optimization with fewer semantic and runtime changes. This PR favors maximum automatic concurrency and is intentionally a more intrusive design prototype.

## Testing

- `cargo +nightly fmt --all -- --check`
- component integration tests, including an exact event log proving parent, child, and grandchild components overlap while output stays in tree order
- dedicated loop test proving every iteration enters before any exits while HTML remains in iterator order
- grammar lowering test across `if`, `match`, loops, and nested markup
- `cargo test --workspace --all-features --locked`

## AI disclosure

OpenAI Codex (GPT-5) helped implement, test, and prepare this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3194,6 +3194,7 @@ name = "topcoat-view"
 version = "0.5.0"
 dependencies = [
  "criterion",
+ "futures-util",
  "http",
  "itoa",
  "memchr",
--- a/crates/topcoat-view/Cargo.toml
+++ b/crates/topcoat-view/Cargo.toml
@@ -18,6 +18,7 @@ http = [
 [dependencies]
 topcoat-core.workspace = true
 
+futures-util.workspace = true
 http = { workspace = true, optional = true }
 itoa.workspace = true
 memchr.workspace = true
--- a/crates/topcoat-view/grammar/src/view/component.rs
+++ b/crates/topcoat-view/grammar/src/view/component.rs
@@ -8,11 +8,14 @@ use syn::{
     spanned::Spanned,
     token::Paren,
 };
-use topcoat_core_grammar::{ParseOption, paths::topcoat_view};
+use topcoat_core_grammar::{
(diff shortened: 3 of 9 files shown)


pikaju:
Thank you Pete, I took a different approach and rewrote the entire rendering system in #317

PR #295, opened 2026-08-03
feat(view): stream deferred views

petehunt:
Depends on #292 and reworks the #290 streaming-shell concept around the ordinary `View` type. Review the commit after `a02c230` for the deferred-view diff.

## Summary

- Add `defer component() { placeholder }` syntax and `View::defer` for manual futures.
- Store deferred work as a view part, so components, layouts, and nested deferred views compose normally without `ShellView`, `include()`, or `finish()`.
- Stream the shell first and completed fragments in completion order; dropping the response body cancels pending work.
- Send inert annotated `<template>` patches and apply them as they arrive with the parser-blocking external `defer_script()` asset, with no inline scripts.
- Add focused coverage, formatter/docs updates, and a runnable `deferred-view` example.

This is more integrated than #290: it adds a deferred view-part variant, lets the router retain an owned request context while streaming, and makes `futures-util` a normal router dependency. In exchange, every view can participate and composition needs no parallel type or helper. Deferred tasks are one-shot; cloned views share the same task.

## Testing

- `cargo +nightly fmt --all`
- `cargo topcoat fmt` (expected Leptos parse errors only)
- `cargo check -p deferred-view`
- `node --check crates/topcoat/browser/defer.js`

Created by OpenAI Codex (GPT-5), which implemented and locally verified the change under maintainer direction.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -665,6 +665,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "deferred-view"
+version = "0.5.0"
+dependencies = [
+ "tokio",
+ "topcoat",
+]
+
+[[package]]
 name = "deranged"
 version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3194,6 +3194,7 @@ name = "topcoat-view"
 version = "0.5.0"
 dependencies = [
  "criterion",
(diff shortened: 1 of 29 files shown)


pikaju:
Thank you for bringing this up, I think it was important to tackle streaming SSR and I underestimated how impactful it would be for the design. I think the #373 approach is the most promising so far, so let's go with that for now.

PR #297, opened 2026-08-03
fix(router): treat empty optional URL-encoded values as none

### LINKED ISSUE #187: Using `Option<jiff::civil::Date>` in a form gives bad request.
Experimented with adding date to `NewTodo` in `toasty-todos` example, specifically as a field `due: Option<Date>`. In response get:
```
bad request: invalid form value: expected four digit year (or leading sign for six digit year), but found end of input (at `due`)
```

This appears to be simply because the blank date selector is being encoded as an empty string, and hence as `Some(...)`, but then jiff fails to parse `""`.

Simple workaround is to use `String` and parse the date in the handler function, but that's not ergonomically great for what I would expect is a common use case. It does mean this can be considered pretty low priority, though.

DeclanRiedel:
## Summary

- Treat empty URL-encoded form and query values as `None` when the target field is an `Option`, including `Option<String>`.
- Preserve existing behavior for required empty strings, map keys, non-empty scalar values, and path-aware parse errors.
- Share the deserializer between `Form<T>` and `#[query_params]`, with regression tests and public documentation for both entry points.

## Design note

This chooses the deserializer-level approach discussed in #187. It intentionally changes an empty value from `Some("")` to `None` for `Option<String>`, for both forms and query parameters.

The issue does not yet record a maintainer preference between this global behavior and an opt-in field helper. This PR remains a draft pending confirmation of that API decision.

## Testing

- `cargo test -p topcoat-router --lib --all-features`
- `cargo clippy -p topcoat-router --all-targets --all-features --locked -- -D warnings`
- `cargo test --workspace --all-features --exclude ui --exclude tailwind --exclude storefront-topcoat`
- `cargo clippy --workspace --all-targets --all-features --locked --exclude ui --exclude tailwind --exclude storefront-topcoat -- -D warnings`
- `RUSTDOCFLAGS="--cfg docsrs -Dwarnings" cargo +nightly doc --workspace --all-features --no-deps --locked --exclude ui --exclude tailwind --exclude storefront-topcoat`
- The three excluded packages execute a downloaded Tailwind binary that cannot run on this NixOS host without `nix-ld`.
- `cargo +nightly fmt --package topcoat-router`
- `cargo topcoat fmt` on the changed Rust files
- Manual HTTP reproduction against `f267417` and the fixed worktree

Closes #187

AI disclosure: OpenAI Codex (`gpt-5.6-sol`) was used to reproduce the issue, implement the deserializer, add tests and documentation, and run local verification.

### DIFF
--- a/crates/topcoat-router/macro/docs/query_params.md
+++ b/crates/topcoat-router/macro/docs/query_params.md
@@ -12,7 +12,9 @@ struct PageQuery {
 
 # Reading the value
 
-[`query_params::<T>(cx)`](fn.query_params.html) parses the current request's query string with [`serde_urlencoded`](https://docs.rs/serde_urlencoded/latest/serde_urlencoded/) and returns `Result<&T, &QueryParamsError>`: a reference to the parsed struct, or to the [`QueryParamsError`](type.QueryParamsError.html) naming the key that failed. Unlike a path parameter, the struct is not tied to a route: any handler can read it. Parsing runs at most once per request; the result is then memoized.
+[`query_params::<T>(cx)`](fn.query_params.html) parses the current request's URL-encoded query string and returns `Result<&T, &QueryParamsError>`: a reference to the parsed struct, or to the [`QueryParamsError`](type.QueryParamsError.html) naming the key that failed. Unlike a path parameter, the struct is not tied to a route: any handler can read it. Parsing runs at most once per request; the result is then memoized.
+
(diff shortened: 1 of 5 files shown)


pikaju:
I can't tell if this is complete, considering it has been in Draft mode for a while, so I will close for now. If you or someone else still needs this, please include reasoning in the description as to why this change makes sense, because it isn't immediately obvious to me

PR #332, opened 2026-08-10
design(view): streaming SSR support

pikaju:
[Rendered](https://github.com/tokio-rs/topcoat/blob/ssr-proposal/DESIGN.md)

[Rendered 2](https://github.com/tokio-rs/topcoat/blob/1d8e31db0d9f484e778c13a3a0def59b74892de4/DESIGN-2.md)

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,258 @@
+# Streaming SSR
+
+This document proposes streaming server-side rendering for Topcoat: a way for a page to send meaningful HTML immediately, render skeleton UI in place of slow data, and swap in the real content over the same HTTP response as it becomes ready.
+
+The design rests on two new primitives. `defer` marks a piece of data as allowed to arrive after the first paint, and `boundary` marks a region of the page as independently swappable. Everything else, including error handling, stays exactly what it is in Topcoat today: plain Rust control flow over plain Rust `Result`s.
+
+## Background
+
+### How Topcoat renders today
+
+A `#[page]` is an async function returning `Result<View>`. Layouts wrap pages by path prefix and receive the inner content as `slot: Result<View>`, so a layout can propagate an error with `slot?` or match on it and render custom error UI. Errors are ordinary Rust values: a component deep in the tree fails with `?`, the error bubbles up through the call stack, and whichever layer wants to handle it handles it. There is one error mechanism and it is the language's:
+
+```rust
+use topcoat::{
+    Result,
+    context::Cx,
+    router::{StatusCode, error::{NotFoundError, RouterErrorExt}, layout, page},
+    view::view,
+};
+
+#[page("/posts/{id}")]
+async fn post(cx: &Cx) -> Result {
+    // `None` becomes a `NotFoundError` and bubbles out with `?`.
+    let post = find_post(cx).await.ok_or_not_found()?;
+    view! { <h1>(&post.title)</h1> }
+}
+
+#[layout("/")]
+async fn root_layout(slot: Result) -> Result {
+    // The error keeps its type on the way out, so the layout can catch it
+    // and replace it with branded UI, or pass anything else along with `?`.
+    let content = match slot {
+        Err(error) if error.downcast_ref::<NotFoundError>().is_some() => view! {
+            (StatusCode::NOT_FOUND)
+            <h1>"Nothing here"</h1>
+        },
+        content => content,
+    }?;
+
+    view! {
+        <html>
+            <body>(content)</body>
+        </html>
+    }
+}
+```
+
+Components inside a `view!` render concurrently. Sibling components, loop iterations, and nested components all start at the same time, so a component waiting on a database query does not delay its siblings and request waterfalls do not happen. What the framework does not do today is respond before everything has finished: the whole view is rendered into one buffer and sent as one response. The response is exactly as slow as the slowest component on the page.
+
+Two recent pieces of infrastructure matter for this proposal:
+
+- The component identity system gives every component invocation a stable 128-bit identity derived from the chain of invocation sites leading down to it, disambiguated inside loops by the `key:` argument. The same invocation reached the same way hashes to the same identity on every render.
+- `#[memoize]` caches a function's result for the duration of a request, keyed by its arguments, and concurrent callers share one in-flight future. Expensive work called from several places in one request runs once.
+
+### The problem
+
+Some data is slow: a search backend, a third-party API, a heavy aggregate query. Today that slowness sits on the critical path of the first byte. The user stares at a blank tab until the slowest query resolves, even though the shell of the page, the navigation, and most of the content were ready long ago.
+
+The well-known fix is streaming: send the fast parts immediately with skeleton placeholders, keep the connection open, and stream the slow parts in as they resolve. React popularized this model with Suspense, so that is the natural starting point.
+
+## Why a Suspense Port Does Not Fit
+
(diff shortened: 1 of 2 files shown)


DESIGN.md
```diff
@@ -0,0 +1,258 @@
...
+                }
+            </div>
+        },
+    }
+}
+```
+
+There is no fallback parameter, no lazy closure, and no new control flow construct. `Deferred` is a plain enum handled with a plain `match`, in the component body or inside `view!`, and both arms are ordinary code with full access to the surrounding scope.
+
+A `defer` call is identified across renders by the component identity of the enclosing body combined with the call site, obtained via `#[track_caller]`. The identity system's existing rules apply unchanged: a `defer` reached through an unkeyed repeated invocation has an ambiguous identity and fails with the error message naming the invocation that needs a `key:`. A `defer` call that itself repeats inside a loop within one component body needs its own key; the API should offer a keyed variant for that case.
```
carllerche, line 136: identifying the call site w/ #[track_caller] means if there is a fn in the middle, it won't work. You could imagine `fn foo(...) { defer(...) }` and *every* component calls `foo`. If you need `defer` to be only callable from a `#[component]` body, you probably need other use a macro or something like that.

DESIGN.md
```diff
@@ -0,0 +1,258 @@
...
+
+Two things are different mid-stream, and both are inherent to streaming rather than to this design:
+
+- The status code and headers went out with the first chunk. Status codes and headers declared by later passes are discarded; an error that bubbles to the very top can no longer turn the response into a 500. The framework renders its error view and swaps it in as a whole-page update. Whole-page error swaps are the unhandled fallback, not the intended path; a page that defers fallible work should let a layout catch the error and render something sensible.
+- Redirects must keep working after the stream has begun. A redirect error surfacing on a later pass is translated into a swap-stream instruction that makes the client navigate, instead of a `Location` header.
+
+### The `boundary` Primitive
```
carllerche, line 166: Can `defer` be an implicit boundary so you can not think about `boundary` in the common case, unless you specifically want to handle errors at a higher level?

DESIGN.md
```diff
@@ -0,0 +1,258 @@
...
+
+Boundary diffing adds a softer expectation: renders should be deterministic, because a boundary that renders differently from the same data hashes differently and gets re-sent and re-swapped for nothing. Freshly generated random ids, timestamps rendered mid-page, or iteration over unordered maps cause spurious swaps. The result is correct but wasteful, and a swap replaces DOM, which discards focus, scroll position, and input state inside the region. For now this is a documentation concern: the user should keep boundary content stable. Tooling, such as a dev-mode warning when a boundary's hash changes although no `defer` inside it changed state, can come later.
+
+## Open Questions
+
+**Future storage and re-execution.** On every pass, the code path leading to a `defer` runs again and constructs a fresh future; the framework must not run the work twice, and must produce the value again on each subsequent pass. Candidate answers, not mutually exclusive:
```
carllerche, line 220: The entire design feels very much like how `async fn` works internally anyway, modulo some differences. You probably want to take some type that can render w/ a `asynchronous fn &mut self` at the root and call it multiple times

DESIGN.md
```diff
@@ -0,0 +1,258 @@
...
+
+The memoize combination is the most attractive because it reuses existing machinery and matches how expensive work should be written anyway, but the exact signature and bounds of `defer` need prototyping.
+
+**Pass scheduling.** Re-rendering once per completed future is wasteful when several complete close together. A short batching window before starting a pass would coalesce them. Whether to batch, and for how long, is undecided.
+
+**Limits.** A page that keeps discovering new `defer` calls streams forever. A cap on passes or a deadline per request, after which pending regions keep their skeletons and the stream closes, is probably wanted.
+
+**Clients without the swap script.** Crawlers and JS-less clients receive the skeleton document and inert templates. The semantics of `defer` permit a degenerate blocking implementation, awaiting the future inline and returning `Ready` on the first pass, which is byte-equivalent to not streaming at all. Offering that as a per-request mode, for example for known bots or for tests, is cheap insurance. The same blocking behavior is the natural fallback for contexts that cannot stream, such as views rendered to strings or mail bodies.
```
carllerche, line 232: Is this not known at request time w/ some header? If it is, you could just not stream (wait for all defers to finish) before responding.

carllerche:
@pikaju I'm trying to capture some thoughts real fast before going to bed. I can try expanding more later.

First, I still think the design proposal is really promising, and at a high level we should keep exploring it. I think there are some specific issues we need to iron out, but if we can do that, there are a lot of very nice things that fall out of it.

I am still very worried about the potential foot guns the specifics of this proposal adds, specifically w.r.t. making it easy to accidentally add slow net calls in places that might be repeatedly called. I know we have some ideas for how to catch those at runtime, but what early Topcoat users tell us is that they really like the **compile-time** checks that Topcoat makes. We really need to try to find the right balance between perf, DX, and compile-time checks.

Thinking more about this, it seems that, fundamentally, a component breaks down into two parts:
* Data-loading, which is asynchronous and should happen *once* per component call.
* Rendering, which is sync, and can happen **N** times per component call, based on `defer` usage.

@petehunt proposed the following in Discord:

```rust
#[component(fetch = load_user)]
fn user_profile(cx: &Cx, user: User) -> Result {
    view! {
        <h1>(user.name)</h1>
    }
}

async fn load_user(
    cx: &Cx,
    user_id: UserId,
) -> Result<User> {
    User::get(cx, user_id).await
}
```

This gets to the point that there is the data load & rendering split. I know you don't like arg providers, and I also am not a fan of the added boilerplate for the common / simple path. It did get me thinking if we could split the two phases with less boilerplate.

I think we should reconsider having the final expression in a component fn be "special". Specifically, we could try extracting the final expression to be the "sync" rendering expression that might be called repeatedly. So:

```rust
#[component]
async fn user_profile(cx: &Cx) -> Result {
    let user = User::get(user_id(cx)).await?;

    view! {
        <h1>(user.name)</h1>
    }
}
```

would compile in a way that splits the last expression (the `view!` call) into something different that could be called many times. I wonder if it could be as simple as wrapping the last expression with a `move || { ... }` but I am not sure what the internals are like right now.

However, I think doing this would a) minimize boilerplate b) capture the patterns you want c) get us compile time enforcing of "no async calls in the render phase".

You can still have regular rust error propagation with `?` and you can still have arbitrary rust code to build up views. Your two options are to slam it in the final expression using `{ ... }` or to just call a sync render func as your last expr:

```rust
#[component]
async fn user_profile(cx: &Cx) -> Result {
    let user = User::get(user_id(cx)).await?;

    if user.admin {
        view! {
            <h1>ADMIN: (user.name)</h1>
        }
    } else {
        view! {
            <h1>(user.name)</h1>
        }
    }
}
```
```rust
#[component]
async fn user_profile(cx: &Cx) -> Result {
    let user = User::get(user_id(cx)).await?;

    render(user)
}
```

I'm not 100% sure how `defer` would work as I'm not super familiar with the internals, but it could be that you build up a deferred loader during data-loading and use it in the render phase:

```rust

#[component]
async fn drink_grid(cx: &Cx) -> Result {
    let drinks = defer(cx, drinks(cx));

    match drinks.poll() {
        Deferred::Pending => view! {
            <div class="grid gap-4 sm:grid-cols-2">
                for _ in 0..6 {
                    <div class="h-32 animate-pulse rounded-lg bg-muted"></div>
                }
            </div>
        },
        Deferred::Ready(drinks) => view! {
            <div class="grid gap-4 sm:grid-cols-2">
                for drink in drinks? {
                    drink_card(key: &drink.slug, drink: drink)
                }
            </div>
        },
    }
}
```

So, it meets the "no async in final expr" goal.

PR #352, opened 2026-08-18
fix(runtime): support await expressions in ExprBlock and ExprIf

### LINKED ISSUE #349: Procedure function can be called only in top level, not in child blocks
## Problem

When procedure function is placed in child blocks, not in top level, inside the closure of `@click`, the Rust code compiles successfully but browser omits an error and it fails to construct the event handler.

## Reproduction

The minimal working code is as follows.
```Rust
use topcoat::{
    asset::{AssetBundle, RouterBuilderAssetExt},
    context::Cx,
    cookie::{Cookie, Cookies, Key, RouterBuilderCookieExt, cookies},
    runtime::{Event, procedure, shard},
    tailwind,
    view::view,
};

#[tokio::main]
async fn main() -> std::io::Result<()> {
    topcoat::start(
        Router::builder()
            .assets(AssetBundle::load()?)
            .discover()
            .cookies()
            .app_context(Key::generate())
            .build(),
    )
    .await
}

#[page("/")]
async fn home(cx: &Cx) -> topcoat::Result {
    view! {
        <!DOCTYPE html>
        <html>
            <head>
                <title>"Hello world"</title>
                topcoat::dev::script()
                topcoat::runtime::script()
                <link rel="stylesheet" href=(tailwind::stylesheet!())>
            </head>
            <body>
                signal username_input = "".to_string();
                signal user = match cookies(cx).get("username") {
                    Some(c) => c.value().to_string(),
                    None => "Guest".to_string(),
                };
                <div class="flex flex-col space-y-2 m-4">
                    current_user(user: $(user.get()))
                    <div class="flex space-x-1">
                        <label for="username">"Username: "</label>
                        <input
                            id="username"
                            type="text"
                            :value=$(username_input.get())
                            @change=$(|e: Event| username_input.set(e.target.value))
                            class="border rounded px-1"
                        />
                    </div>
                    <button
                        @click=$(async |_e| {
                            let username = username_input.get();
                            // XXX: procedure can be called only in top level, not in if block!!
                            // if !username.is_empty() {
                                set_username_to_cookie(username.clone()).await;
                                user.set(username);
                                username_input.set("".to_owned());
                            // }
                        })
                        class="cursor-pointer rounded px-2 py-1 bg-green-500 text-white"
                    >
                        "Set username"
                    </button>
                </div>
            </body>
        </html>
    }
}

#[procedure]
async fn set_username_to_cookie(cx: &Cx, name: String) -> topcoat::Result<&'static str> {
    let cookie = cookies(cx);
    cookie.add(Cookie::build(("username", name)).path("/").build());

    Ok("ok")
}

#[shard]
async fn current_user(user: String) -> topcoat::Result {
    view! {
        <p>
            "Current user: "
            (user)
        </p>
    }
}

```

When `if !username.is_empty()` block in the closure of `@click` is commented out, Rust still compiles successfully but on the browser the following error is shown in the console.

```
VM1833:3 Uncaught SyntaxError: Unexpected identifier '__external1' (at VM1833:3:417)
    at new Function (<anonymous>)
    at Et (topcoat-f897074dd8132591.js:1:14173)
    at Jt (topcoat-f897074dd8132591.js:1:17088)
    at K (topcoat-f897074dd8132591.js:1:16978)
    at U.start (topcoat-f897074dd8132591.js:1:17971)
    at topcoat-f897074dd8132591.js:1:18011
Et	@	topcoat-f897074dd8132591.js:1
Jt	@	topcoat-f897074dd8132591.js:1
K	@	topcoat-f897074dd8132591.js:1
start	@	topcoat-f897074dd8132591.js:1
(anonymous)	@	topcoat-f897074dd8132591.js:1
```
The error message is a bit cryptic and I could not find an apparent problem at the specified parts in `topcoat-f897074dd8132591.js`.

Using the procedure in a simple `{}` block instead of `if` one also fails.
When `set_username_to_cookie(username.clone()).await;` line is commented out while the if block is active, the error goes away.

## Root cause
Above all, I speculate the problem is trying to call procedure function NOT in top level inside the closure of `@onclick`.
If this behavior is not an intended one, it would be nice if it is allowed to write procedures also in child blocks, empowering to write more complex logics on event handlers.

RoseWrightdev:
fixes https://github.com/tokio-rs/topcoat/issues/349

In cases like:

```rust
@click=$(async |_e| {
    if true {
        let _v = test_procedure("hello test".to_owned()).await;
    }
}
```

or 

```rust
@click=$(async |_e| {
    {
        let _v = test_procedure("hello test".to_owned()).await;
    }
}
```

The runtime would never wrap the anonymous js function with async, even if the node had an async child. Now we walk the tree and check if there is an async child before assigning it sync or async.
 
AI Disclosure: claude helped me find the root cause because this was my first time using `syn`


EbaraKoji:
This change seems to solve the issue #349 .

However, the tests might to be incomplete because it does not check the js syntax of created function.
For example, when `js.push_str(if is_async { ")())" } else { ")()" });` line is commented out in `expr-block.rs`, the test `procedure_call_inside_block_is_an_async_func` still passes but the created js funtion on the browser causes SyntaxError because of not closing brace, as expected.
In addition, in the test `procedure_call_inside_if_is_an_async_func`, `asset!` part is the same as test of block's and actual if logics are not considered.

In this PR the `expr_block` and `expr_if` appears OK at first glance, and I understand checking js function's syntax completely in Rust tests is very difficult. I think merging this PR with recognizing these problems is a valid option.

But when also trying to add await expressions in such as `ExprLoop`, `ExprWhile`, and `ExprForLoop`, things could be more complex and I think it's better if there're basic js syntax evaluations and logic checks.

Anyway, I think enabling await expressions in this way is a big progress.

pikaju (dismissed):

RoseWrightdev:
> This change seems to solve the issue #349 .
> 
> However, the tests might to be incomplete because it does not check the js syntax of created function. For example, when `js.push_str(if is_async { ")())" } else { ")()" });` line is commented out in `expr-block.rs`, the test `procedure_call_inside_block_is_an_async_func` still passes but the created js funtion on the browser causes SyntaxError because of not closing brace, as expected. In addition, in the test `procedure_call_inside_if_is_an_async_func`, `asset!` part is the same as test of block's and actual if logics are not considered.
> 
> In this PR the `expr_block` and `expr_if` appears OK at first glance, and I understand checking js function's syntax completely in Rust tests is very difficult. I think merging this PR with recognizing these problems is a valid option.
> 
> But when also trying to add await expressions in such as `ExprLoop`, `ExprWhile`, and `ExprForLoop`, things could be more complex and I think it's better if there're basic js syntax evaluations and logic checks.
> 
> Anyway, I think enabling await expressions in this way is a big progress.

I agree that there should be better testing on the javascript side, ideally something like a headless cypress run, but that's a bigger architectural repo-wide change than I'm ready to take on.

pikaju (approved):

PR #354, opened 2026-08-18
fix(router): rewrite through 'discover' routing

ieuanmeredith:
Upon testing **request rewrites** in v0.6.0 I was seeing persistent 'internal server error' instead of the desired behaviour.

The culprit was `PageWithLayouts::handle -> slot.into_response` being called on `RewriteError`. The submitted change allows the `RewriteError` back into the router.rs loop that correctly handles the rewrite.

Tests have been added to `router.rs` to confirm the fix for;
1. pages
2. layouts
3. pages through layouts.

The error example has been updated to show a RewriteError working as expected.

Tests were generated with GPT-5.6 Terra.

### DIFF
--- a/crates/topcoat-router/src/page.rs
+++ b/crates/topcoat-router/src/page.rs
@@ -4,7 +4,7 @@ use topcoat_core::{context::Cx, error::Result};
 use topcoat_view::View;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


pikaju (dismissed):

pikaju:
Ah, good catch... Put too much trust in the clanker there. Thank you!

PR #359, opened 2026-08-18
fix(view): keep if-let bindings alive in concurrent renders

### LINKED ISSUE #358: captured values don't live long enough when passing to sub-components after upgrading to 0.6.1
After the changes in 0.6.x, I'm now getting lifetime errors within a view telling me that arguments are not living long enough after being captured by a coroutine:

```
error[E0597]: `status` does not live long enough
   --> web/src/components.rs:360:52
    |
353 | /     view! {
354 | |         <div>
355 | |             for item in items {
356 | |                 <div>
...   |
359 | |                             if let Some(status) = item.user_status {
    | |                                         ------ binding `status` declared here
360 | |                                 badge_test(status: status)
    | |                                                    ^^^^^^ borrowed value does not live long enough
...   |
368 | |         </div>
369 | |     }
    | |     -
    | |     |
    | |_____`status` dropped here while still borrowed
    |       value captured here by coroutine
```

Minimal reproducer:
```rust
#[derive(Debug, Clone, Copy)]
pub enum Status {
    First,
    Second,
}

pub struct Obj {
    name: String,
    user_status: Option<Status>,
    post_status: Option<Status>,
}

#[component]
pub async fn test(items: &[&Obj]) -> topcoat::Result {
    view! {
        <div>
            for item in items {
                <div>
                    <span>(&item.name)</span>
                    <div>
                        if let Some(status) = item.user_status {
                            badge_test(status: status)
                        }
                        if let Some(status) = item.post_status {
                            badge_test(status: status)
                        }
                    </div>
                </div>
            }
        </div>
    }
}

#[component]
pub async fn badge_test(status: Status, #[default] mut attrs: Attributes) -> topcoat::Result {
    let (klass, text) = match status {
        Status::First => ("first", "First"),
        Status::Second => ("second", "Second"),
    };
    view! {
        <div class=(class!(klass, "badge", attrs.remove("class"))) (attrs)>(text)</div>
    }
}

```

Note that if I only have one of those `if let Some(status)...` blocks, I get no errors, but if I have both of them, I get lifetime errors for both sub-component call sites. Cloning the status variable doesn't seem to make any difference, and it's hard to know what's happening in the background within the view! macro to know what might be going wrong.

mannlohchab:
## Summary

- Joined `if let` / `match` arms in `view!` wrap their bodies in `async` blocks so sibling components can render concurrently. Pattern bindings from those arms were captured by reference and dropped when the arm returned the future, which is the E0597 after 0.6 concurrent rendering (two `if let Some(status)` component calls failed; one did not).
- Stash those bindings in `Option` cells on the enclosing hoist and rebind them inside the branch future so the coroutine owns them for as long as it runs.
- Unit variants such as `None` are not treated as bindings.

Closes #358

## Test plan

- [x] `cargo +nightly fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- [x] `cargo test --workspace --all-features`
- [x] `RUSTDOCFLAGS="--cfg docsrs -Dwarnings" cargo +nightly doc --workspace --all-features --no-deps --locked`
- [x] `cargo check -p topcoat --target wasm32-unknown-unknown --no-default-features --features router,view,runtime,asset,cookie,session,discover,multipart,tower --locked`
- [ ] Confirm the original two-`if let` component call site compiles

### DIFF
--- a/crates/topcoat-view/grammar/src/view/hir/emit.rs
+++ b/crates/topcoat-view/grammar/src/view/hir/emit.rs
@@ -1,6 +1,6 @@
 use proc_macro2::{Span, TokenStream};
 use quote::{TokenStreamExt, format_ident, quote, quote_spanned};
-use syn::Ident;
+use syn::{Expr, Ident, Pat, Token};
 use topcoat_core_grammar::paths::topcoat_view;
 
 /// HIR nodes that emit themselves into the two phases of an [`Emitter`].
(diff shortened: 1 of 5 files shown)


pikaju:
Thank you I appreciate it, but I implemented a similar solution in #360

PR #363, opened 2026-08-19
feat(cli): expose run method on `TopcoatCli`

kyrias:
Making this public allows for embedding the Topcoat CLI in other applications, for example using the [xtask] pattern.

  [xtask]: https://github.com/matklad/cargo-xtask


pikaju (approved):
Thank you!

PR #364, opened 2026-08-20
feat(cli): port retry on dev server init

RoseWrightdev:
When using the CLI, I kept running into the problem where the port was occupied. Most CLIs ask you to try another port if this happens, but as it is now `topcoat dev` crashes. I made it so it selects the first port. :) Open to feedback on styling here, and I also get if this needs to be thrown out because the whole CLI will be rewritten soon.

Screenshot 2026-08-20 at 7 50 14 PM


pikaju (approved):
Thank you. I changed it to be automatic so you don't have to approve running on a different port. I think that's less tedious

PR #365, opened 2026-08-21
fix(cli): don't reload over in-flight navigations on reconnect

### LINKED ISSUE #331: `topcoat dev` sometimes seems to refuse to serve slow pages
I made a change in my web application that seemed to introduce a slow database query on a certain page (let's call it page 'B'). Page 'A' contains a link to page 'B'. Previously, when running under the dev server, I could click this link in my browser and page 'A' would open. But after my change, when I click the link to page 'B', that page never opens, and instead page 'A' just seems to get reloaded. 

When I look at the 'network' tab in my web inspector when I click the link, I see the GET request to page B, which never completes. Then I see a GET request to ws://127.0.0.1:59039/ws, and then I see another GET request to page 'A'. This results in page 'A' just reloading and never navigating to page B. I can no longer navigate to page 'B' when running my app with the topcoat dev server. If I run it with `cargo run`, things work fine (albeit a bit slowly). To be clear, I intend to fix the slowness affecting page B, but I still expect to be able to navigate to this page. I'm not entirely sure how to debug this farther, hopefully you have some ideas what's happening. For the record, the slow page is on the order of 0.5s response time.

ankarhem:
The browser closes the page's dev WebSocket during navigation - at navigation start in Firefox, at commit in Chrome. reconnect() treated any close as the dev server going away and reloaded as soon as a probe connected, which always succeeds because the broadcast server outlives app rebuilds. Firing ~500ms into a still-pending navigation, that reload cancelled it and snapped the browser back to the old URL.

Track pending navigations via the Navigation API and defer probing while one is in flight; navigateerror resumes probing when a navigation fails and the page survives, and pageshow resets the flag for pages restored from the back/forward cache.

Reproduction / verification repo (swap flake input to fix branch): https://github.com/ankarhem/topcoat-ws-unload

Assisted by: glm-5.3

Closes #331

### DIFF
--- a/crates/topcoat-cli/src/dev/dev.js
+++ b/crates/topcoat-cli/src/dev/dev.js
@@ -203,10 +203,25 @@
     ws.onclose = reconnect;
   }
 
+  // The browser also closes the WebSocket when the page navigates away - at
+  // navigation start in Firefox, at commit in Chrome. Reloading then would
+  // cancel the in-flight navigation, so probing is deferred while one is
+  // pending; navigateerror resumes it when a navigation fails and the page
+  // lives on.
+  let navigating = false;
+  window.navigation?.addEventListener("navigate", () => (navigating = true));
+  window.navigation?.addEventListener(
+    "navigateerror",
+    () => (navigating = false)
+  );
+  // A page restored from the back/forward cache is no longer navigating.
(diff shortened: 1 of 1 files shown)


pikaju (approved):
Thank you!

PR #375, opened 2026-08-28
fix(runtime): preserve boolean procedure results

elliottminns:
## Problem

I ran into an issue where a Procedure returning a `Result<bool>` resolved to `undefined` in a browser runtime expression.

For example:

```rust
#[procedure]
async fn enabled() -> Result<bool> {
      Ok(true)
}

view! {
    signal enabled = false;

    <button @click=$(async |_e| {
        let result = enabled().await; // result becomes undefined instead of true
        enabled.set(result);
    })>
        "Check"
    </button>

    <p>"Enabled: " $(enabled.get())</p>
   }
}
```

The procedure response contains the correct `true` value, but the result becomes `undefined` after crossing the JavaScript Promise boundary.

## Cause

The browser `BoolSurrogate` implements a method named `then` to mirror Rust's `bool::then` method.

JavaScript treats any object with a callable `then` property as a Promise-like thenable. This means that when a hydrated `Bool` is returned from a procedure, Promise resolution invokes `Bool.then` and resolves it without the boolean value, causing it to become `undefined`.

## Fix

Rename the browser surrogate method from `then` to `then_` and compile Rust calls to `bool::then` as JavaScript calls to `then_`.

This preserves the existing Rust API: `condition.then(|| value)` whilst ensuring hydrated boolean values are not treated as JavaScript thenables.

## Verification

Added regression coverage that:

- Awaits a Bool through a Promise boundary and verifies the value is preserved.
- Verifies the browser surrogate still supports Rust `bool::then` behavior through `then_`.
- Verifies Rust `.then(...)` expressions compile to JavaScript `.then_(...)`.

Commands run:

```
corepack yarn test
cargo test -p topcoat-runtime-grammar -p topcoat-runtime-macro -p topcoat-runtime
cargo check
```

All tests pass.

## AI assistance

OpenAI Codex assisted with reproducing the browser behavior, identifying the JavaScript thenable collision, implementing the fix, and adding regression tests. I reviewed and verified the resulting changes manually.


pikaju (approved):
Thank you!

PR #382, opened 2026-09-05
docs: fix "hello world" examples so that they compile

andrew-rj:
Hi there! While walking through the `getting_started.md`, I noticed that these examples did not compile. I've updated them minimally so that they compile.


pikaju:
I think you mixed up the versions here, the getting started guide is already written for Topcoat 0.7, which will be released shortly.

PR #383, opened 2026-09-05
fix(router): treat an empty form or query value as None for Option<T>

### LINKED ISSUE #187: Using `Option<jiff::civil::Date>` in a form gives bad request.

MavenRain:
## Summary

A browser submits a blank input as `due=` (key present, value empty).  `Form<T>` and `#[query_params]` handed that empty value to the field's own parser, so any `Option<T>` whose `T` rejects `""` (`f64`, `bool`, `jiff::civil::Date`, an enum, ...) failed with `bad request: invalid form value: ...`.  Only `Option<String>` survived, as `Some("")`.

This adds a small in-crate deserializer (`crates/topcoat-router/src/urlencoded.rs`) that mirrors `serde_urlencoded` with one rule added: an empty value deserializes to `None` for any `Option<T>`.  `Form<T>` and `#[query_params]` now decode through it.  Serialization keeps using `serde_urlencoded`, and `QueryParamsError` is the same type as before.

Behavior changes, each covered by a test:
- `Option<T>` from an empty value is `None` for every `T`, including enums and other types with their own parser (was a bad request).
- `Option<String>` from an empty value is `None` (was `Some("")`).
- Unchanged: a required `String` reads `""`, a required number still returns a bad request, and a missing key for `Option<T>` is `None`.

Alternative considered: switching to `serde_html_form`, the fork `axum-extra` uses.  It only maps an empty value to `None` for a fixed list of primitive types (`bool`, integers, floats), so the `Option<Date>` from the issue would still fail.  If you would rather keep `Some("")` for `Option<String>`, I can drop that part of the rule.

Closes #187

## Testing

- `cargo test -p topcoat-router --lib`: 445 passed, 0 failed (Rust 1.98.1)
- `cargo clippy -p topcoat-router --no-deps --all-targets -- -D warnings`: clean
- `cargo fmt -p topcoat-router -- --check`: clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2810,7 +2810,6 @@ dependencies = [
  "inventory",
  "serde",
  "serde_json",
- "serde_urlencoded",
  "tokio",
  "tokio-tungstenite",
  "topcoat-alpine-ajax",
--- a/crates/topcoat-router/macro/docs/query_params.md
+++ b/crates/topcoat-router/macro/docs/query_params.md
@@ -12,7 +12,7 @@ struct PageQuery {
 
 # Reading the value
 
-[`query_params::<T>(cx)`](fn.query_params.html) parses the current request's query string with [`serde_urlencoded`](https://docs.rs/serde_urlencoded/latest/serde_urlencoded/) and returns `Result<&T, &QueryParamsError>`: a reference to the parsed struct, or to the [`QueryParamsError`](type.QueryParamsError.html) naming the key that failed. Unlike a path parameter, the struct is not tied to a route: any handler can read it. Parsing runs at most once per request; the result is then memoized.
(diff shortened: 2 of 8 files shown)


pikaju:
Thank you!

pikaju (approved):

PR #396, opened 2026-09-09
feat(core)!: make Error cheap to clone

y-l-g:
## Summary

Makes `topcoat_core::Error` cheaply cloneable by storing the error in an `Arc` instead of using `anyhow::Error`.

## Motivation

`#[memoize(as_ref)]` caches `Result<&T, &Error>`, so a memoized fallible loader receives a *borrowed* error where its signature needs an *owned* one:

```rust
#[memoize(as_ref)]
async fn load_users(cx: &Cx) -> Result<Vec<User>> { /* ... */ }

async fn users(cx: &Cx) -> Result<&Vec<User>> {
    // Before: the borrowed error can only be rebuilt from its message —
    // type and source chain are lost.
    // load_users(cx).await.map_err(|e| std::io::Error::other(e.to_string()).into())

    // After: clone the cached error, type and chain preserved.
    load_users(cx).await.map_err(Error::clone)
}
```

### DIFF
--- a/crates/topcoat-core/src/error.rs
+++ b/crates/topcoat-core/src/error.rs
@@ -1,6 +1,7 @@
 use std::{
     fmt::{Debug, Display},
-    ops::{Deref, DerefMut},
+    ops::Deref,
+    sync::Arc,
 };
 
 pub type Result<T = (), E = Error> = ::core::result::Result<T, E>;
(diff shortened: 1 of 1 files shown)


pikaju:
Thank you for the idea @y-l-g , I think Arc is the way to go, I made a bunch of extra changes to round this off

pikaju (approved):

jonner:
After quickly trying out the new release, it looks like topcoat::Error no longer implements `std::error::Error` which caused breakage in my project related to `thiserror`. I had this in my project:
```rust
#[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error(transparent)]
     Topcoat(#[from] topcoat::Error),
    // more stuff
}
```

After upgrading I get the following error
```
error[E0599]: the method `as_dyn_error` exists for reference `&topcoat::Error`, but its trait bounds were not satisfied
  --> web/src/error.rs:3:13
   |
 3 |     #[error(transparent)]
   |             ^^^^^^^^^^^ method cannot be called on `&topcoat::Error` due to unsatisfied trait bounds
   |
  ::: /home/x/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/topcoat-core-0.9.0/src/error.rs:21:1
   |
21 | pub struct Error(Arc<dyn ErrorObject>);
   | ---------------- doesn't satisfy `_: AsDynError<'_>` or `topcoat::Error: StdError`
   |
   = note: the following trait bounds were not satisfied:
           `topcoat::Error: StdError`
           which is required by `topcoat::Error: thiserror::__private21::AsDynError<'_>`
           `&topcoat::Error: StdError`
           which is required by `&topcoat::Error: thiserror::__private21::AsDynError<'_>`

For more information about this error, try `rustc --explain E0599`.
```

It no longer allows me to use the `transparent` error message or the use of `#[from]` in the error variant definition. Is this intentional?

Issue #394, opened 2026-09-09
Improve Tower service integration for prefixed routes


Martell0x1:
## Problem

When using a Tower service such as `tower_http::services::ServeDir` with `TowerRoute`, the route prefix is not stripped from the request path before it is passed to the Tower service.

For example:

```rust
use tower_http::services::ServeDir;

Router::builder()
    .discover()
    .route(TowerRoute::new(
        Methods::Any,
        "/res/{*path}",
        ServeDir::new("res"),
    ))
    .build()
```

Given the following directory structure:

```text
res/
└── hello.txt
```

a request to:

```text
GET /res/hello.txt
```

matches the Topcoat route, but the Tower service still receives the `/res/hello.txt` path.

As a result, `ServeDir::new("res")` may resolve the file as:

```text
res/res/hello.txt
```

instead of:

```text
res/hello.txt
```

This makes mounting a Tower file-serving service under a route prefix unintuitive and differs from the behavior users may expect from APIs such as Axum's `nest_service`.

## Expected behavior

When a Tower service is mounted under a route prefix such as `/res`, the prefix should either be stripped before passing the request to the service, or Topcoat should provide a convenient abstraction that handles this behavior.

For example:

```text
/res/hello.txt
       │
       │ strip /res
       ▼
/hello.txt
       │
       ▼
ServeDir::new("res")
       │
       ▼
res/hello.txt
```

## Possible solution

Investigate improving the Tower integration so that route prefixes can be handled correctly when passing requests to Tower services.

One possible approach could be a built-in file-serving route handler with behavior similar to `ServeDir`, while another could be handling path-prefix stripping within `TowerRoute` or through a dedicated Tower integration.

The exact API and implementation should be discussed.

## Motivation

Serving static files from a local directory is a common requirement for web applications. Making this work naturally with Topcoat's routing and Tower integration would make migration from existing Axum/Tower applications easier and provide a more ergonomic way to serve static assets.

Martell0x1:
i am working on this

PR #397, opened 2026-09-09
fix(router): strip prefix layer

Martell0x1:


pikaju:
Hey @Martell0x1, thank you for the PR. I think we should have a more general solution, something like a layer that strips a prefix that you can then apply to your `ServeDir` tower route. I don't think we should have special treatment for `ServeDir` like this. It might also be nice to have our own serve dir support so you don't need tower at all, and our own handler could be smart enough to strip the prefix by itself.

Martell0x1:
> Hey @Martell0x1, thank you for the PR. I think we should have a more general solution, something like a layer that strips a prefix that you can then apply to your `ServeDir` tower route. I don't think we should have special treatment for `ServeDir` like this. It might also be nice to have our own serve dir support so you don't need tower at all, and our own handler could be smart enough to strip the prefix by itself.

Thanks for the review. Dropped the `ServeDir` special case in `TowerRoute`.

`StripPrefix` is a normal router layer: it rewrites the request URI for inner layers and the route, keeps the query string, and leaves the original URI on the outer context. Apply it to any tower service that needs paths relative to its mount, including tower-http's `ServeDir`:

```rust
Router::builder()
    .route(TowerRoute::new(Methods::Any, "/res/{*path}", ServeDir::new("res")))
    .layer(StripPrefix::new("/res"))
    .build();
```

There is also a native `ServeDir` route (behind `serve`) so you do not need tower for static files. It strips its own route prefix, serves `GET`/`HEAD`, maps directories to `index.html`, and 404s on path traversal.

`cargo test -p topcoat-router --all-features` is green.

pikaju:
@Martell0x1 thank you, I refactored some things and removed the `ServeDir` because I added `DirectoryRoute` in another PR. So I believe this now fixes #394

Martell0x1:
> @Martell0x1 thank you, I refactored some things and removed the `ServeDir` because I added `DirectoryRoute` in another PR. So I believe this now fixes #394

Great , Thanks

